PR #12, opened 2023-12-11
Add katex header & bump libs version

survived:
Fixes the build failure on docs.rs. Releases a new version of the lib `v0.1.1`

### DIFF
--- a/generic-ec-core/Cargo.toml
+++ b/generic-ec-core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "generic-ec-core"
-version = "0.1.1"
+version = "0.1.2"
 edition = "2021"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/dfns/generic-ec"
--- a/generic-ec-core/Cargo.toml
+++ b/generic-ec-core/Cargo.toml
@@ -13,7 +13,7 @@ generic-array = "0.14"
 subtle = { version = "2.4", default-features = false }
 rand_core = { version = "0.6", default-features = false }
 zeroize = { version = "1", default-features = false }
-serde = { version = "1", default-features = false, optional = true }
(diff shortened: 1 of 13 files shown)


generic-ec-core/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "generic-ec-core"
-version = "0.1.0"
+version = "0.1.2"
```
maurges, line 3: Why are this and generic-ec bumped to 1.2?

survived: because I managed to fix the issue only after second attempt

survived: you can see the git history

maurges: I see, it's because the versions were published to crates.io. I would vote for bumping all versions to 1.2 then

survived: Why bumping what has not been changed? If they were supposed to have the same version number, they would be a single lib

maurges: But nothing changed in all 4 crates, according to semver. We could even yank the old version and republish with the same number.

I thought they are different crates to ease development and namespacing, but they are still interconnected. Sitting here thinking about it, I can see weak technical arguments for simultaneous and not bumping of versions, and my strongest feeling here is aesthetics which also doesn't matter much. So I leave it to your choice.

maurges (approved):

generic-ec-core/Cargo.toml
survived: If I could reupload library with the same version I'd do this, but it's impossible. Yanking doesn't delete crate version, it just marks it "do not use me", which is not even mandatory.

survived: I don't see strong arguments in favor of bumping all 4 crates, so I leave them as they are

PR #13, opened 2024-01-25
Add `ed25519` curve

survived:


generic-ec/Cargo.toml
```diff
@@ -14,7 +14,7 @@ keywords = ["elliptic-curves"]
 
 [dependencies]
 generic-ec-core = "0.1"
-generic-ec-curves = { version = "0.1", optional = true }
+generic-ec-curves = { path = "../generic-ec-curves", optional = true }
```
maurges, line 17: Note: this is what I was talking about when I said it's stupid in cargo that selecting a package and selecting where to get the package are not be separate

maurges (approved):

generic-ec/Cargo.toml
survived: It looks fine to me :thinking:

PR #21, opened 2024-02-28
Implement missing traits

survived:
Implements:
* `Partial{Eq, Ord}` for `NonZero<T>`
* `Sum<NonZero<Point<E>>>` for `Point<E>`
* `Samplable` for `NonZero<SecretScalar<E>>`

### DIFF
--- a/generic-ec-zkp/src/polynomial.rs
+++ b/generic-ec-zkp/src/polynomial.rs
@@ -48,7 +48,7 @@ mod requires_alloc {
         /// let x = Scalar::random(&mut OsRng);
         /// assert_eq!(
         ///     coefs[0] + x * coefs[1] + x * x * coefs[2],
-        ///     polynomial.value(&x),
+        ///     polynomial.value::<_, Scalar<_>>(&x),
         /// );
         /// ```
         pub fn from_coefs(mut coefs: Vec<C>) -> Self {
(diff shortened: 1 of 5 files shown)


generic-ec/src/non_zero/mod.rs
```diff
@@ -296,6 +316,57 @@ impl<E: Curve> AsRef<Scalar<E>> for NonZero<SecretScalar<E>> {
     }
 }
 
+impl<T> cmp::PartialEq<T> for NonZero<T>
```
maurges, line 312: We don't have `PartialEq<Self> for NonZero<T>`, or do we?

generic-ec-zkp/src/polynomial.rs
```diff
@@ -400,7 +400,10 @@ mod tests {
         // 2. Commit to the secret
         // F(x) = G * f(x)
         let F = &f * &Point::generator();
-        assert_eq!(&secret * Point::generator(), F.value(&Scalar::zero()));
+        assert_eq!(
+            &secret * Point::generator(),
+            F.value::<_, Point<_>>(&Scalar::zero())
+        );
```
maurges, line 406: Why did it change, can we fix it back? I hate explicit types

generic-ec/src/non_zero/mod.rs
survived: We do, it's autoderived

generic-ec-zkp/src/polynomial.rs
survived: I think it's because I implemented PartialEq for NonZero, and now when compiler sees `Point<E>` being compared with some generic `O`, it can't anymore deduce that `O = Point<E>` because it could be `NonZero<Point<E>>` as well

maurges (approved):

PR #22, opened 2024-02-28
Publish next version

maurges:
Bump versions for release. Generic-ec-zkp had commits in tests, so it's not getting releases.

Fix generic-ec-curves not demanding a feature of zeroize

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -146,8 +146,9 @@ jobs:
         # when it's published
         # - generic-ec-curves
         - generic-ec-zkp
+    steps:
     - uses: actions/checkout@v3
     - name: Check semver
       uses: obi1kenobi/cargo-semver-checks-action@v2
       with:
-        package: ${{ package }}
+        package: ${{ matrix.package }}
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -134,3 +134,20 @@ jobs:
         cache-on-failure: "true"
     - name: Dry-run publish
       run: cargo publish --dry-run -p ${{ matrix.package }}
+  check-semver:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        package:
+        - generic-ec
+        - generic-ec-core
+        # TODO: this fails because the currently published version doesn't compile
+        # by itself. A fix is present and will be published soon, uncomment this
+        # when it's published
+        # - generic-ec-curves
+        - generic-ec-zkp
+    - uses: actions/checkout@v3
+    - name: Check semver
+      uses: obi1kenobi/cargo-semver-checks-action@v2
+      with:
+        package: ${{ package }}
--- a/generic-ec-curves/Cargo.toml
+++ b/generic-ec-curves/Cargo.toml
@@ -13,7 +13,7 @@ generic-ec-core = { version = "0.1", path = "../generic-ec-core", default-featur
 
 subtle = { version = "2.4", default-features = false }
 rand_core = { version = "0.6", default-features = false }
-zeroize = { version = "1", default-features = false }
+zeroize = { version = "1", default-features = false, features = ["zeroize_derive"] }
 
 crypto-bigint = { version = "0.5", default-features = false, optional = true }
 elliptic-curve = { version = "0.13", default-features = false, features = ["sec1", "hash2curve"], optional = true }
--- a/generic-ec-curves/Cargo.toml
+++ b/generic-ec-curves/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "generic-ec-curves"
-version = "0.1.3"
+version = "0.1.4"
 edition = "2021"
 license = "MIT OR Apache-2.0"
 description = "Elliptic curves for `generic-ec` crate"
--- a/generic-ec/Cargo.toml
+++ b/generic-ec/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "generic-ec"
-version = "0.1.5"
+version = "0.2.0"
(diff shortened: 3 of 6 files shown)


generic-ec-curves/Cargo.toml
```diff
@@ -13,7 +13,7 @@ generic-ec-core = { version = "0.1", path = "../generic-ec-core", default-featur
 
 subtle = { version = "2.4", default-features = false }
 rand_core = { version = "0.6", default-features = false }
-zeroize = { version = "1", default-features = false }
+zeroize = { version = "1", default-features = false, features = ["zeroize_derive"] }
```
survived, line 16: I think the feature is just called `derive`

survived: see their [Cargo.toml](https://github.com/RustCrypto/utils/blob/82f5b1bae8e8744c1a2538a73d5e9dd6a64a5b62/zeroize/Cargo.toml#L28)

generic-ec/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "generic-ec"
-version = "0.1.4"
+version = "0.1.5"
```
survived: The fact that we had to do write explicit generics in `polynomial.value()` suggests that it's probably a breaking change since it does break existing code

generic-ec-curves/Cargo.toml
maurges: https://docs.rs/zeroize/latest/zeroize/derive.Zeroize.html - still called by the full name

generic-ec/Cargo.toml
maurges: Semver check didn't find it =) But I agree actually. Polynomial is part of generic-ec-zkp though?

survived: Semver only looks for the common issues, we shouldn't rely on it to find all possible issues

survived: The problem is that before the changes, if the client code (let it be generic-ec-zkp or any other lib) had comparison like `Point<E> == O` where `O` is some type to be deduced by the compiler, compiler previously could figure out that `O = Point<E>`, since point could only be compared with another point, but now it breaks compilation

survived: I guess, both `generic-ec` and `zkp` need major version bump, `generic-ec` because of the changes, and `zkp` because it exposes `generic-ec` types in public API

generic-ec-curves/Cargo.toml
survived: Yeah, and they are also suggesting to use `zeroize_derive` in the docs (hand-written part of the docs). `zeroize` is nicer and shorter tho. My guess that they added an alias but just didn't update the docs

maurges: Seems so. It was two years ago: https://github.com/RustCrypto/utils/pull/712/files

I don't care which one to use

survived:
For some reason, after you moved a semver check into `rust.yml`, no job from `rust.yml` is even started

.github/workflows/rust.yml
```diff
@@ -134,3 +134,20 @@ jobs:
...
+        # TODO: this fails because the currently published version doesn't compile
+        # by itself. A fix is present and will be published soon, uncomment this
+        # when it's published
+        # - generic-ec-curves
+        - generic-ec-zkp
+    - uses: actions/checkout@v3
+    - name: Check semver
+      uses: obi1kenobi/cargo-semver-checks-action@v2
+      with:
+        package: ${{ package }}
```
survived: I think it should be `${{ matrix.package }}` like above

generic-ec-curves/Cargo.toml
survived: Me neither, I'm fine with this one

.github/workflows/rust.yml
maurges: We'll never know because the builds are still stuck, even after pushing 3 commits

.github/workflows/rust.yml
```diff
@@ -134,3 +134,21 @@ jobs:
...
+        # by itself. A fix is present and will be published soon, uncomment this
+        # when it's published
+        # - generic-ec-curves
+        - generic-ec-zkp
+    steps:
+    - uses: actions/checkout@v3
+    - name: Check semver
+      uses: obi1kenobi/cargo-semver-checks-action@v2
+      with:
+        package: ${{ matrix.package }}
```
survived, line 154: It won't fail the pipeline if the version wasn't changed at all? E.g. we don't publish a new version of `core`, so we didn't bump its version, but we still do the semver check, I wonder if it could be a problem

maurges: Yeah, I have the same question. Can you try it locally?

survived: Do we have any published crates that didn't change?

survived: Ah never mind, I can just go back in git history

survived: Seems to work fine
```
$ cargo semver-checks -p generic-ec
     Parsing generic-ec v0.1.4 (current)
      Parsed [  12.163s] (current)
     Parsing generic-ec v0.1.4 (baseline)
      Parsed [  12.944s] (baseline)
    Checking generic-ec v0.1.4 -> v0.1.4 (no change)
     Checked [   0.013s] 62 checks; 62 passed, 0 unnecessary
    Finished [  25.144s] generic-ec
```

survived: Btw I had no problems with installing it, I just had to add cmake to nix-shell

survived (approved):

PR #28, opened 2024-03-25
Add `generic_ec::serde::PreferCompact`

survived:
Note: I couldn't make `Compact` to accept both compact and non-compact formats, as there are some limitations to parsing the ambiguous data, specifically, `PreferCompact` can't accept list-encoded structs (see the `PreferCompact` docs)

### DIFF
--- a/tests/tests/serde.rs
+++ b/tests/tests/serde.rs
@@ -14,7 +14,7 @@ mod tests {
             let point_compressed = point.to_bytes(true).to_vec().leak();
             let point_compressed_hex = hex::encode(&point_compressed).leak();
 
-            // Human-readable, uncompressed
+            // Human-readable
             serde_test::assert_ser_tokens(
                 &point.readable(),
                 &[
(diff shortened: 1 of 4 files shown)


tests/tests/serde.rs
```diff
@@ -148,14 +157,210 @@ mod tests {
...
+                    Token::Seq { len: None },
+                ];
+                tokens.extend(point_uncompressed.iter().copied().map(Token::U8));
+                tokens.extend([Token::SeqEnd, Token::SeqEnd]);
+
+                serde_test::assert_de_tokens(&point.readable(), &tokens);
+            }
+
+            // PreferCompact doesn't support seq-encoded structs
+            serde_test::assert_de_tokens_error::<serde_test::Compact<PreferCompact<Point<E>>>>(
```
maurges, line 198: I don't know how serde tests work, but why is PreferCompact embedded in Compact here?

survived: serde_test::Comact says that data should be deserialized with `human_readable = false` (as opposite to `serde_test::Readable`)

survived: if deserialization logic relies on `is_human_readable` flag, but the type isn't wrapped into either `serde_test::{Compact, Readable}`, assertion panics

maurges (approved):
Knowing that the tests pass in https://github.com/dfns/cggmp21/pull/96 gives me security in these changes

PR #29, opened 2024-03-28
Multiscalar multiplication w/o precomputations

survived:
We often have computations like:

$$Q = s_1 P_1 + \dots + s_n P_n$$

which can be computed more efficiently than calculating each $s_i P_i$ separately and $\sum$-ing them.

In this PR, I implement various techniques for computing multiscalar multiplication efficiently. 

**Note:** docs fail to check in CI because some dependency fails to compile on nightly


generic-ec-core/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "generic-ec-core"
-version = "0.1.3"
+version = "0.1.4"
```
maurges, line 3: Need to bump the required version in `generic-ec`, since you use the new function

survived: I always forget to do that...

survived: Updated

maurges: I don't see the update in generic-ec/Cargo.toml

generic-ec/src/multiscalar.rs
```diff
@@ -0,0 +1,406 @@
...
+//!
+//! Typically, we're working with scalars of 256 bits size. You can easily calculate
+//! that, for instance, even for $n = 3$, the total cost is $768 D + 387 A$. However,
+//! for the same $n$, Straus algorithm has $\text{cost} = 300 D + 192 A$ which
+//! is already significantly faster than naive algorithm.
+//!
+//! You can see the perfromance comparison for different algorithms on secp256k1 curve
+//! on the plot below:
+//!
+//! <img src="https://raw.githubusercontent.com/dfns/generic-ec/m/perf/multiscalar/secp256k1.svg"/>
```
maurges: Probably better to use explicit commit in the link

survived: But then I will forget to update it if it gets changed

generic-ec-core/Cargo.toml
survived: I forgot to push it. Done

generic-ec/src/multiscalar.rs
maurges: Screenshot 2024-04-05 at 12 09 00

I mean

maurges: I want to say that even if you update it, do you really want the old docs to use the new graph? Also I'm pretty sure in the future we will move those pictures around and the docs will break.

Also, here's one alternative solution: https://users.rust-lang.org/t/include-images-in-rustdoc-output/3487/4

maurges: Another solution: since it's an svg, you can just insert it literally or as base64 into the docstring

generic-ec/src/multiscalar.rs
```diff
@@ -0,0 +1,406 @@
...
+            .map(|point_i| {
+                iter::successors(Some(*point_i), |point| Some(point + point_i))
+                    .take(15)
+                    .collect::<Vec<_>>()
+            })
+            .collect::<Vec<_>>();
+
+        // `serialized_len` is amount of radix256-digits. We multiply it by 2 and get
+        // amount of radix16-digits
+        let num_digits = 2 * Scalar::<E>::serialized_len();
```
maurges: Alternative: implement [ExactSizeIterator](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html) for Radix16Iter

generic-ec/src/multiscalar.rs
```diff
@@ -0,0 +1,406 @@
...
+
+        // table[i] = [point_i, 2 * point_i, ..., 15 * point_i]
+        let table = points
+            .iter()
+            .map(|point_i| {
+                iter::successors(Some(*point_i), |point| Some(point + point_i))
+                    .take(15)
+                    .collect::<Vec<_>>()
+            })
+            .collect::<Vec<_>>();
```
maurges: Since it's a matrix, we could get rid of Vec inside Vec and allocate contiguously to improve performace. But it's somewhat inconvenient to write

generic-ec/src/multiscalar.rs
```diff
@@ -0,0 +1,406 @@
...
+                })
+                .sum::<Point<E>>();
+
+            if j == 0 {
+                sum = partial_sum
+            } else {
+                // sum = 16 * sum + partial_sum
+                let sum_at_16 = sum.double().double().double().double();
+                sum = sum_at_16 + partial_sum;
+            }
```
maurges, line 261: Medium nitpick: I think it would be somewhat easier to understand if `sum = if j == 0 { sum } else { sum.double().double().double().double() };` was at the top of the loop, similar to how it's usually done with polynomial values for example.

generic-ec/src/multiscalar.rs
survived: > Note that this trait is a safe trait and as such does not and cannot guarantee that the returned length is correct

generic-ec/src/multiscalar.rs
survived: I think it would be more confusing. It should be obvious that we shift `sum` by 3 bits so we could write our result. If we don't do that when we write our result, this is confusing

generic-ec/src/multiscalar.rs
survived: Ah I thought this comment was referring to another thing. Yes we can do that. Since it would be our implementation, we can trust that it guarantees that the returned length is correct

generic-ec/src/multiscalar.rs
survived: I changed it to include svg into the docs

generic-ec/src/multiscalar.rs
survived: Actually, having `scalars[0].len()` could be confusing as it could lead reader to thinking that different scalars could have different length... We can probably create a static method on `Radix16Iter`, but I don't know what it would be called

generic-ec/src/multiscalar.rs
survived: Good point! Converted it into the one vec

generic-ec/src/multiscalar.rs
maurges: 4 bits you mean? Ok I see your point, let's leave as is

generic-ec/src/multiscalar.rs
```diff
@@ -0,0 +1,406 @@
...
+            return Point::zero();
+        }
+
+        // table[i * 15 .. (i + 1) * 15] = [point_i, 2 * point_i, ..., 15 * point_i]
+        let table = points
+            .iter()
+            .flat_map(|point_i| {
+                iter::successors(Some(*point_i), |point| Some(point + point_i))
+                    .take(15)
+                    .collect::<Vec<_>>()
```
maurges: Can be removed, flat_map accepts iterators directly

survived: Oops, forgot about it

generic-ec/src/multiscalar.rs
survived: I implemented [ExactSizeIterator](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html) for Radix16Iter as you suggested, see if it looks fine

generic-ec/src/multiscalar.rs
```diff
@@ -0,0 +1,404 @@
...
+            return Point::zero();
+        }
+
+        // `serialized_len` is amount of radix256-digits. We multiply it by 2 and get
+        // amount of radix16-digits
+        let num_digits = 2 * Scalar::<E>::serialized_len();
+
+        let mut result = Point::zero();
+        for i in 0..num_digits {
+            let mut buckets = [None; 15];
```
maurges, line 352: Why not start with zeroes and get rid of matching below?

survived: Because I want to minimize additions. Setting it to Point::zero() would add 1 extra addition per each bucket

maurges: After thinking about it, you might be right, especially if points are CT; and if they are not CT it should be comparable

maurges: I actually benchmarked it for n=150 and 100, and the difference is 8% in favour of branching. Ok

generic-ec/src/scalar.rs
```diff
@@ -366,3 +421,62 @@ impl<E: Curve> udigest::Digestable for Scalar<E> {
...
+            is_be,
+            next_radix16: None,
+            next_index: 0,
+        }
+    }
+}
+
+impl<E: Curve> Iterator for Radix16Iter<E> {
+    type Item = u8;
+
```
maurges, line 452: > When implementing an ExactSizeIterator, you must also implement [Iterator](https://doc.rust-lang.org/std/iter/trait.Iterator.html). When doing so, the implementation of [Iterator::size_hint](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint) must return the exact size of the iterator.

https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html

survived: Updated

generic-ec/src/multiscalar.rs
```diff
@@ -0,0 +1,404 @@
...
+#[cfg(feature = "alloc")]
+impl<E: Curve> MultiscalarMul<E> for Default {
+    fn multiscalar_mul<S, P>(scalar_points: impl IntoIterator<Item = (S, P)>) -> Point<E>
+    where
+        S: AsRef<Scalar<E>>,
+        P: AsRef<Point<E>>,
+    {
+        let (mut scalars, points): (Vec<Radix16Iter<E>>, Vec<Point<E>>) = scalar_points
+            .into_iter()
+            .map(|(scalar, point)| (scalar.as_ref().as_radix16_be(), *point.as_ref()))
```
maurges, line 105: Kind if stupid, but how do we guarantee that scalars are all the same length? I don't see us padding them with zeroes, and I don't see the iterator impl adding zeroes too

survived: `Scalar::to_{be|le}_bytes` always returns scalar zero-padded if needed

survived: And `Scalar::as_radix16_{be|le}` uses `Scalar::to_{be|le}_bytes` internally

maurges: I see. That's not obvious from the `Bytes` type, but ok

maurges (approved):

PR #30, opened 2024-04-17
Faster multiscalar multiplication

survived:
* Rewrite `Straus` to work with scalars in NAF
* Remove Pippenger as it's less efficient than the new version of Straus on all `n`
* Change `MultiscalarMult` trait interface. We now require that size of the iterator is known.


generic-ec/src/multiscalar.rs
```diff
@@ -136,269 +110,43 @@ impl<E: Curve> MultiscalarMul<E> for Naive {
...
-    where
-        S: AsRef<Scalar<E>>,
-        P: AsRef<Point<E>>,
-    {
-        let (mut scalars, points): (Vec<Radix16Iter<E>>, Vec<Point<E>>) = scalar_points
+        let (scalars, points): (Vec<_>, Vec<_>) = scalar_points
             .into_iter()
-            .map(|(scalar, point)| (scalar.as_ref().as_radix16_be(), *point.as_ref()))
+            .map(|(s, p)| (s.as_ref().as_raw().0, p.as_ref().as_raw().0))
             .unzip();
```
maurges: You can avoid having two vecs here:

```rust
        let scalar_points = scalar_points.into_iter().collect::<Vec<_>>();
        let scalars = scalar_points.iter().map(|(s, _)| &s.as_ref().as_raw().0);
        let points = scalar_points.iter().map(|(_, p)| &p.as_ref().as_raw().0);
```

If we restrict the iterator to be Clone, we could avoid vecs at all, but I don't think it's viable sadly

generic-ec/src/multiscalar/straus.rs
```diff
@@ -0,0 +1,327 @@
...
+            .zip(iter::once(true).chain(iter::repeat(false)))
+        {
+            if !is_first_iter {
+                r = r.double();
+            }
+            for (naf, lookup_table) in nafs.iter().zip(&lookup_tables) {
+                let naf_i = naf[i];
+                match naf_i.cmp(&0) {
+                    core::cmp::Ordering::Greater => {
+                        r += lookup_table.get(naf_i as usize);
```
maurges: Suggestion: `naf_i.unsigned_abs().into()`

maurges: To distinguish between the narrowing conversion below and a safe conversion here

generic-ec/src/multiscalar/straus.rs
```diff
@@ -0,0 +1,327 @@
...
+    fn iter(&self) -> impl Iterator<Item = &[i8]> {
+        self.matrix.chunks_exact(self.naf_size)
+    }
+}
+
+/// Read one or more u64s stored as little endian bytes.
+///
+/// ## Panics
+/// Panics if `src.len() != 8 * dst.len()`.
+fn read_le_u64_into(src: &[u8], dst: &mut [u64]) {
```
maurges, line 250: Suggestion: read either le or be bytes depending on platform. I also tried to write a transmute with unsafe casts, but since alignment is not guaranteed, in the worst case we would have to copy the whole slice anyway.

maurges:
```
// Safety: safe if `src.as_ptr()` is aligned to u64
unsafe fn transmute_aligned_native_u64(src: &[u8]) -> &[u64] {
    let size = std::mem::size_of::<u64>();
    assert!(src.len() % size == 0);
    // Safety:
    // 1. a) Data is valid for reads of len * size bytes as the original data is
    //    valid for exactly this
    //    b) Data is non-null - as original slice
    //    c) Data properly aligned for u64 - guaranteed by caller
    // 2. Data points to initialized values of type u64, since any 8 bytes are a
    //    proper u64
    // 3. Memory is not mutated at all
    // 4. Total size is correct as it's gotten from the original size
    unsafe {
        std::slice::from_raw_parts(src.as_ptr() as *const u64, src.len() / size)
    }
}
```

generic-ec/src/multiscalar/straus.rs
```diff
@@ -0,0 +1,327 @@
...
+                // This window bits are contained in a single u64
+                (x_u64[u64_idx] >> bit_idx) & self.window_mask
+            } else {
+                // Combine the current u64's bits with the bits from the next u64
+                ((x_u64[u64_idx] >> bit_idx) | (x_u64[u64_idx + 1] << (64 - bit_idx)))
+                    & self.window_mask
+            };
+
+            // Add the carry into the current window
+            let window = if carry { bit_buf + 1 } else { bit_buf };
```
maurges, line 214: `let window = bit_buf + u64::from(carry);` - though both are ok

generic-ec/src/multiscalar/straus.rs
survived: I would prefer not writing unsafe code when possible. Does this optimization has noticeable effect on performance?

generic-ec/src/multiscalar/straus.rs
survived: That's what I wrote at first but then I thought that checking bool could be cheaper than converting it to u64 and adding to the bit_buf

generic-ec/src/multiscalar/straus.rs
survived: Nice one!

generic-ec/src/multiscalar/straus.rs
survived: So do I correctly understand that you're suggesting to do `scalar.to_le_bytes()` or `scalar.to_be_bytes()` depending on the platform and then do `transmute_aligned_native_u64`? Doesn't that reorder u64 on different platforms? If `u64`s are in different order on different platforms, that would additionally complicate the NAF conversion function.

survived: Also, did you notice that we always have a trailing zero u64 in the end of vec of u64? Which means we would anyway need to allocate memory to have this trailing `0_u64`

survived: It's probably possible to rewrite the algorithm so it doesn't require trailing zero, but it's again an additional complication to the algorithm

maurges: > Also, did you notice that we always have a trailing zero u64 in the end of vec of u64 

I did after I wrote my option =) Yeah, that might eat away the gains from transmuting.

What I suggest at the minumum is to have two functions: `read_le_u64_into` and `read_be_u64_into`, and we serialize the scalar to correct byte order and dispatch between the two based on the platform

survived: I'm confused, what would be advantage of having two functions? You think converting be bytes to u64 on be platform is noticeably faster?

generic-ec/src/multiscalar/straus.rs
maurges: I'm pretty sure they both get optimized to the same `test carry carry; repz inc`, it's more a question of style

generic-ec/src/multiscalar/straus.rs
maurges: Yes it should be faster? Because converting from be to be is a noop, and converting from be to le requires shuffling bytes around

survived: I don't really like this idea because this complicates the algorithm (now we need to serialize scalars in be/le depending on the platform, then rearrange `u64`s, or change the algorithm so we accept native order, but output little-endian), and then we can't test it (we probably can set up CI with both little-endian and big-endian instances, but that's additional work to do)

survived: I don't think the profit is worth it

maurges: Ok, I agree for testability

maurges (approved):
I think I get hof NAF construction works. But don't ask me this in a month because it will take me another day to remember

PR #33, opened 2024-04-24
fix latex '\text' grammar

cx19981114:


survived:
Why? It renders fine on docs.rs

cx19981114:
I find some information regarding the LaTeX amsmath package(https://www.ams.org/arc/tex/amsmath/amsldoc.pdf). 
1. The '\text'  command for typesetting a fragment of text inside a display. 
2. In all multiline environments, lines are divided by '\\' (I entered double backslashes here, but it's showing it as just one. :)

survived:
We're using [katex](https://katex.org/) for rendering latex expressions in web, it might behave differently from amsmath package. It looks like there's no issue on the web with displaying expressions, everything looks normal in the docs (e.g. this is how [Polynomial::from_coefs](https://docs.rs/generic-ec-zkp/latest/generic_ec_zkp/polynomial/struct.Polynomial.html#method.from_coefs) is rendered, it seems fine before the changes). Right now, I don't see what issue the PR is addressing, so I'm closing it. If you find anything that doesn't render properly, please, point it out to us (e.g. send a screenshot).

PR #34, opened 2024-04-25
More efficient impl for `Scalar::from_{be|le}_bytes_mod_order`

### LINKED ISSUE #31: Optimize `Scalar::from_bytes_{le|be}_mod_order`
`Scalar::from_{le|be}_bytes_mod_order` are implemented generically for any scalar using scalar arithmetic. However, curve library might have a more optimized conversion function. For instance, `curve25519_dalek` has a functions `from_bytes_mod_order` and `from_bytes_mod_order_wide` which are significantly faster.

```
ed25519/512/generic-ec  time:   [13.040 µs 13.048 µs 13.057 µs]
ed25519/512/dalek::from_bytes_mod_order_wide
                        time:   [103.11 ns 103.23 ns 103.44 ns]
```

We should leverage this backend optimization.

survived:
Closes #31


survived:
@maurges PR is ready to review, but I'm going push more code, I want to benchmark the curves implementations

generic-ec-core/src/lib.rs
```diff
@@ -185,3 +238,13 @@ impl<N: ArrayLength<u8>> ByteArray for GenericArray<u8, N> {
...
+
+/// Reduces an integer represented as array of `N` bytes modulo curve (prime) order
+pub trait Reduce<const N: usize> {
+    /// Interprets `bytes` as big-endian encoding of an integer, returns this
+    /// integer modulo curve (prime) order
+    fn from_be_array_mod_order(bytes: &[u8; N]) -> Self;
+    /// Interprets `bytes` as little-endian encoding of an integer, returns this
+    /// integer modulo curve (prime) order
+    fn from_le_array_mod_order(bytes: &[u8; N]) -> Self;
+}
```
maurges, line 250: This got me to think: why do we have GenericArray in this project at all instead of using const generics? Do you remember what's stopping us?

generic-ec-curves/src/utils.rs
```diff
@@ -0,0 +1,270 @@
...
+            padded[..len].copy_from_slice(bytes);
+            S::from_le_array_mod_order(&padded)
+        }
+        64 => {
+            #[allow(clippy::expect_used)]
+            let bytes: &[u8; 64] = bytes.try_into().expect("we checked that bytes len == 64");
+            S::from_le_array_mod_order(bytes)
+        }
+        65.. => {
+            let two_to_512 = S::add(&S::from_le_array_mod_order(&[0xff; 64]), one);
```
maurges, line 41: Maybe
```
let mut bytes = [0; 65];
bytes[0] = 1;
S::from_le_array_mod_order(&bytes);
```

maurges: Also, isn't this two_to_512?

generic-ec-curves/src/utils.rs
```diff
@@ -0,0 +1,270 @@
...
+                #[allow(clippy::expect_used)]
+                let chunk: &[u8; 32] = chunk.try_into().expect("wrong chunk size");
+                S::from_be_array_mod_order(chunk)
+            });
+
+            remainder
+                .into_iter()
+                .chain(chunks)
+                .reduce(|acc, int| S::add(&S::mul(&acc, &two_to_256), &int))
+                .unwrap_or_default()
```
maurges, line 236: What do you think to refactor this part into a separate function? It differs in 32/64, chunks/rchunks, from_be_array_mod_order/from_le_array_mod_order, and which function to recurse to.

maurges: You could even replace the recursing call with from_le_array_mod_order

generic-ec/src/scalar.rs
```diff
@@ -142,25 +142,14 @@ impl<E: Curve> Scalar<E> {
...
     pub fn from_be_bytes_mod_order(bytes: impl AsRef<[u8]>) -> Self {
-        let scalar_0x100 = Scalar::from(0x100_u16);
-        bytes
-            .as_ref()
-            .iter()
-            .fold(Scalar::<E>::zero(), |acc, byte| {
-                acc * scalar_0x100 + Scalar::from(*byte)
-            })
+        let scalar = E::Scalar::from_be_bytes_mod_order(bytes.as_ref());
+        Self::from_raw(scalar)
```
maurges, line 146: I'm now surprised we ever came up with something so inefficient

maurges: Though I'm also surprised that the bit-by-bit algorithm in multiscalar is efficient at all

tests/tests/curves.rs
```diff
@@ -162,16 +162,24 @@ mod tests {
     fn scalar_from_bytes_mod_order<E: Curve>() {
         let mut rng = DevRng::new();
 
-        let random_scalar = Scalar::<E>::random(&mut rng);
-        for s in [Scalar::zero(), random_scalar] {
-            let s_be = s.to_be_bytes();
-            let s_le = s.to_le_bytes();
+        let fixed_len = [8, 16, 32, 48, 64, 80, 96, 128, 256, 512];
```
maurges: Add at least one odd length?

tests/benches/measure_perf.rs
```diff
@@ -30,11 +30,9 @@ fn multiscalar_for_curve<E: Curve>(
 ) where
     multiscalar::Naive: MultiscalarMul<E>,
     multiscalar::Straus: MultiscalarMul<E>,
-    multiscalar::Pippenger: MultiscalarMul<E>,
```
maurges, line 33: We should add a verification that benchmarks compile to the CI

generic-ec-core/src/lib.rs
survived: Const generics in the traits are very limited. E.g. this code doesn't compile:

```rust
trait Foo {
    const N: usize;
    fn foo(&self, arr: [u8; Self::N]);
}
```

(which we actually need most of the times)

generic-ec-curves/src/utils.rs
survived: It will fall into infinite recursion :smiley:

survived: or I misunderstand the code. `S::from_le_array_mod_order(&bytes)` requires that `Reduce<65>` is implemented for the scalar. There's no scalar which implements it

generic-ec-curves/src/utils.rs
survived: Here we indeed can remove the recursion call, I will do that. But you can't remove recursion in `32_64` call, which shows that we can't move it into separate function and reuse across 32 and 32_64

generic-ec/src/scalar.rs
survived: It was generic enough, there's no uniform support of reducing 256/512-bits integers mod order, which requires to write more code, and it wasn't that obvious that the code is very inefficient until I did benchmarks

generic-ec-curves/src/utils.rs
maurges: Ah right, my bad

generic-ec-curves/src/utils.rs
maurges: Challenge accepted

maurges:
```rust
fn reduce_chunks<'a, const N: usize, F, I, S>(
    is_be: bool,
    remainder: &'a [u8],
    chunks: I,
    from_array: F,
    one: &S,
) -> S
where
    F: Fn(&[u8; N]) -> S,
    I: DoubleEndedIterator<Item = &'a [u8]>,
    S: Default,
    S: generic_ec_core::Additive + generic_ec_core::Multiplicative<S, Output = S>,
{
    let modulo = S::add(&from_array(&[0xff; N]), one);
    let remainder = if !remainder.is_empty() {
        let mut padded = [0u8; N];
        let len = remainder.len();
        if is_be {
            padded[N - len..].copy_from_slice(remainder);
        } else {
            padded[..len].copy_from_slice(remainder);
        }
        Some(from_array(&padded))
    } else {
        None
    };
    let chunks = chunks.rev().map(|chunk| {
        #[allow(clippy::expect_used)]
        let chunk: &[u8; N] = chunk.try_into().expect("wrong chunk size");
        from_array(chunk)
    });

    remainder
        .into_iter()
        .chain(chunks)
        .reduce(|acc, int| S::add(&S::mul(&acc, &modulo), &int))
        .unwrap_or_default()
}
```
used like
```rust
        33.. => {
            let chunks = bytes.rchunks_exact(32);
            reduce_chunks(
                true,
                chunks.remainder(),
                chunks,
                |a: &[u8; 32]| S::from_be_array_mod_order(a),
                one,
            )
        }
```

generic-ec-core/src/lib.rs
maurges: Ok, I actually see one meaningful use of GenericArray, and it's the one that is forbidden similarly to what you show

generic-ec-curves/src/utils.rs
survived: That would only work for `32` tho, it won't work for `32_64`

survived: In `32_64` remainder can fit into either 32 or 64 bytes array. If it fits into 32 bytes array, conversion is cheaper, so we should use that (this is why there's the recursion in the first place)

maurges: You're right, I had a look at ed25519 and 64 is a lot more expensive. But that makes it easier to write:
```rust
// call site
            reduce_chunks(
                chunks.remainder(),
                chunks,
                |a: &[u8; 64]| S::from_le_array_mod_order(a),
                scalar_from_le_bytes_mod_order_reducing_32_64,
                one,
            )
// declaration
fn reduce_chunks<'a, const N: usize, F, G, I, S>(
    remainder: &'a [u8],
    chunks: I,
    from_array: F,
    rec: G, // <-- new
    one: &S,
) -> S
where
    F: Fn(&[u8; N]) -> S,
    G: Fn(&[u8], &S) -> S, // <-- new
    I: DoubleEndedIterator<Item = &'a [u8]>,
    S: Default,
    S: generic_ec_core::Additive + generic_ec_core::Multiplicative<S, Output = S>,
// a tweak
    let remainder = if !remainder.is_empty() {
        Some(rec(remainder, one))
    } else {
        None
    };
```

maurges: I just found out that lifetimes in `impl Fn` are now supported, so you can eve make the declaration 2 (or 3) lines shorter

survived: Imo it's getting too complex, I'd rather leave it as is

maurges: I disagree, but I do think we're not going to touch this code anymore, unless we decide to rewrite it completely, so ok

maurges (dismissed):

tests/src/bin/analyze.rs
```diff
@@ -218,3 +232,125 @@ fn fmt_and_write_svg(path: impl AsRef<std::path::Path>, svg: &str) -> Result<()>
...
+            .build()
+            .with(tabled::settings::style::Style::markdown())
+    );
+
+    Ok(())
+}
+
+fn choose_uniform_unit<'a>(
+    measurements: impl Iterator<Item = &'a Measurement> + Clone,
+) -> (Vec<f64>, &'static str) {
```
maurges, line 305: Why not map back to Measurements, only adjusting their units?

survived: In this way, signature of function guarantees that units are indeed the same

maurges: Alternatively, since measurements are no longer displayed, the Display instance above is not necessary?

survived: Yep I just deleted it (it was buggy anyways)

maurges (approved):

PR #38, opened 2024-05-22
Changelog workflow

maurges:
Add a workflow that checks if we updated the changelog


maurges:
You can see it working in the workflows in the commit history of https://github.com/dfns/generic-ec/pull/37

survived (approved):
Nice!

PR #40, opened 2024-06-06
Add hash to scalar primitive

survived:
<s>Pipeline is failing because `rand_hash` repo is private. Pending approval to make it public.</s> Upd: CI is working


generic-ec/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "generic-ec"
-version = "0.3.0"
+version = "0.4.0-rc1"
```
survived, line 3: Is my judgment right that we need a major bump if we upgrade udigest from v0.1 to v0.2? @maurges

maurges: It's used in public APIs, so yes we should

maurges: Same goes for generic-ec-zkp as I see

maurges:
> Pipeline is failing because rand_hash repo is private

How does it even work for you locally? Do you have some override to fetch from github?

survived:
Yep I have this in my `~/.gitconfig`:

```
[url "ssh://git@github.com/dfns/rand_hash"]
	insteadOf = https://github.com/dfns/rand_hash
```

survived:
@maurges rand_hash was made public, CI is working

maurges:
I think we should publish rand-hash before merging and publishing this one.

generic-ec-zkp/Cargo.toml
```diff
@@ -12,8 +12,8 @@ keywords = ["elliptic-curves", "zk-proof"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-generic-ec = { version = "0.3", path = "../generic-ec", default-features = false }
-udigest = { version = "0.1", features = ["derive"], optional = true }
+generic-ec = { version = "0.4.0-rc1", path = "../generic-ec", default-features = false }
+udigest = { version = "0.2.0-rc1", features = ["derive"], optional = true }
```
maurges, line 16: You should bump generic-ec-zkp version since you're bumping generic-ec dep

survived:
> I think we should publish rand-hash before merging and publishing this one.

rand_hash is published, I updated the code to pull it from crates io

maurges (approved):

PR #41, opened 2024-06-17
Update udigest

survived:
Changelog pipeline is failing, but there's nothing to add to changelog in this case

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,3 @@ members = [
 exclude = [
   "wasm/nostd",
 ]
-
-[patch.crates-io.udigest]
-git = "https://github.com/dfns/udigest"
-branch = "small-improvements"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -11,3 +11,7 @@ members = [
 exclude = [
   "wasm/nostd",
 ]
+
+[patch.crates-io.udigest]
+git = "https://github.com/dfns/udigest"
+branch = "small-improvements"
```
survived: to be deleted

maurges (commented):
Fun, so the changelog script doesn't live well with rc versions

survived:
Yeah, but that doesn't happen very often that we have rc versions, so it's easier to ignore it than to update the script

maurges (dismissed):

survived:
There's another release of `udigest` upcoming... I think it's better to update everything in bulk when udigest `v0.2` is out

survived:
@maurges updated deps so there's no `rc` versions anymore

maurges (approved):

PR #42, opened 2024-08-02
Add contribution & vuln reporting guidelines

survived:


survived:
Clippy errors seem to be false-positive — when I remove references as suggested, it doesn't compile anymore

UPD: it was already reported [here](https://github.com/rust-lang/rust-clippy/issues/13170)

maurges (approved):
After looking through clippy issue tracker, seems like a regression in 1.80.

PR #43, opened 2024-08-22
Fix no_std support after updating to latest udigest

survived:
generic-ec is dependent on `std` feature of `udigest` which breaks no_std support. It used to do no harm as `std` feature in `udigest` was no-op, however, in the latest update it's no longer no-op.

PR removes dependency on `std` feature of `udigest`

Clippy still produces false-positives in the tests

Changelog check also seem to produce a false-positive

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,3 +12,40 @@ exclude = [
   "wasm/nostd",
 ]
 
+[workspace.dependencies]
+generic-array = "0.14"
+
+hex = { version = "0.4", default-features = false }
+
+rand = "0.8"
+rand_core = { version = "0.6", default-features = false }
+
+serde = { version = "1", default-features = false }
+serde_json = "1"
+serde_with = { version = "2", default-features = false }
+
+sha2 = { version = "0.10", default-features = false }
+
+subtle = { version = "2.4", default-features = false }
+
+udigest = { version = "0.2.0", default-features = false }
+
+zeroize = { version = "1", default-features = false }
+
+criterion = "0.5"
+generic-tests = "0.1"
+serde_test = "1"
+rand_dev = "0.1"
+
+[workspace.dependencies.curve25519]
+package = "curve25519-dalek"
+version = "4"
+default-features = false
+
+[patch.crates-io.udigest]
+git = "https://github.com/dfns/udigest"
+branch = "udigest_as"
+
+[patch.crates-io.udigest-derive]
+git = "https://github.com/dfns/udigest"
+branch = "udigest_as"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,11 +41,3 @@ rand_dev = "0.1"
 package = "curve25519-dalek"
 version = "4"
 default-features = false
-
-[patch.crates-io.udigest]
-git = "https://github.com/dfns/udigest"
-branch = "udigest_as"
-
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -12,3 +12,40 @@ exclude = [
...
+version = "4"
+default-features = false
+
+[patch.crates-io.udigest]
+git = "https://github.com/dfns/udigest"
+branch = "udigest_as"
+
+[patch.crates-io.udigest-derive]
+git = "https://github.com/dfns/udigest"
+branch = "udigest_as"
```
survived: to be removed

wasm/nostd/Cargo.toml
```diff
@@ -8,3 +8,11 @@ publish = false
...
 generic-ec-curves = { path = "../../generic-ec-curves", default-features = false }
 generic-ec-zkp = { path = "../../generic-ec-zkp", default-features = false, features = ["udigest"] }
+
+[patch.crates-io.udigest]
+git = "https://github.com/dfns/udigest"
+branch = "udigest_as"
+
+[patch.crates-io.udigest-derive]
+git = "https://github.com/dfns/udigest"
+branch = "udigest_as"
```
survived: to be deleted

Cargo.toml
```diff
@@ -12,3 +12,40 @@ exclude = [
...
+zeroize = { version = "1", default-features = false }
+
+criterion = "0.5"
+generic-tests = "0.1"
+serde_test = "1"
+rand_dev = "0.1"
+
+[workspace.dependencies.curve25519]
+package = "curve25519-dalek"
+version = "4"
```
maurges, line 42: This line is what confuses the changelog detector. Maybe shall we write it in the `{}` style?

survived: It's a workspace cargo toml, it should not be looked at in the first place...

survived: Looking at the script, it only checks the cargo toml when there's changlelog in the same directory, so this cargo toml is not checked

survived:
@maurges all deps are updated, it's ready

generic-ec-curves/Cargo.toml
```diff
@@ -11,30 +11,28 @@ repository = "https://github.com/dfns/generic-ec"
...
 
 group = { version = "0.13", default-features = false, optional = true }
 
 [dependencies.curve25519]
-package = "curve25519-dalek"
-version = "4"
-default-features = false
+workspace = true
 features = ["group", "zeroize", "rand_core", "precomputed-tables"]
 optional = true
```
maurges, line 29: Oops, I meant this line with version is what confuses the script. And it's already removed, so no problems. Still, why not rewrite it in `{}` style? It is my personal nitpick though, I just don't like this part of toml syntax usage

survived: I like it though, it's neat when `{}` style results into too long lines.

survived: or did you already get used to >200 lines? :D

survived: `curve25519 = { workspace = true, features = ["group", "zeroize", "rand_core", "precomputed-tables"], optional = true }`

nah, it's certainly less readable

maurges (approved):

PR #52, opened 2025-01-17
Add curve-specific aliases

survived:


maurges (approved):
generic-ecies inspired? (=

survived:
sort of!

PR #54, opened 2025-01-23
Add dlog_eq zkp to generic-ec-zkp

maurges:


maurges:
I've been writing docs for the past 16 hours, send help

generic-ec-zkp/src/dlog_eq.rs
```diff
@@ -0,0 +1,277 @@
+//! Knowledge of equality of two discrete logarithms $\Pi^\text{dlog_eq}$
+//!
+//! This is a protocol that lets the prover $\P$ convince the
+//! verifier $\V$ that it knows the discrete logarithms of two values, and that
+//! those logarithms are equal. $\P$ knows the secret data $x$ - the discrete
+//! logarithm. $\P$ and $\V$ share common data:
+//! - $H$ - a point on an elliptic curve
+//! - $X = x \cdot G$
+//! - $Z = x \cdot H$
```
survived: It is not necessary that $G$ is (the) generator. In general, common data is:

- Points $H_1, H_2$
- $P_1 = x \cdot H_1$
- $P_2 = x \cdot H_2$

And prover proves that it knows $x_1 = \log_{H_1} P_1$ and $x_2 = \log_{H_2} P_2$ and proves that $x_1 = x_2$

generic-ec-zkp/src/dlog_eq.rs
```diff
@@ -0,0 +1,277 @@
...
+#[derive(Debug, Clone, Copy)]
+#[cfg_attr(feature = "udigest", derive(udigest::Digestable), udigest(bound = ""))]
+pub struct Data<E: Curve> {
+    /// `X`, the value inside `log_G`
+    pub pub_share: Point<E>,
+    /// `H`, the base for the second logarithm
+    pub base: Point<E>,
+    /// `Z`, the value inside the second logarithm `log_H`
+    pub exp: Point<E>,
+}
```
survived: So here, for sake of generality, it's worth adding the second base, and change naming perhaps (why `exp`, we use multiplicative notation!)

```suggestion
pub struct Data<E: Curve> {
    /// Generator $H_1$
    pub base1: Point<E>,
    /// Generator $H_2$
    pub base2: Point<E>,
    /// Point $P_1 = x_1 \cdot H_1$ such that $x_1$ is only known to the prover
    pub point1: Point<E>,
    /// Point $P_2 = x_2 \cdot H_2$ such that $x_2$ is only known to the prover
    pub point2: Point<E>,
}
```

generic-ec-zkp/src/dlog_eq.rs
```diff
@@ -0,0 +1,277 @@
...
+    pub fn commit<E: Curve>(base: Point<E>, r: Scalar<E>) -> Commitment<E> {
+        (Point::generator() * r, base * r)
+    }
+    /// First round of the protocol: commit to well-constructed [`Data`] by
+    /// itself. Produces [`Commitment`] to be sent
+    ///
+    /// - `r` - random nonce. Can be produced by `Scalar::random(&mut rng)`
+    pub fn commit_data<E: Curve>(data: Data<E>, r: Scalar<E>) -> Commitment<E> {
+        commit(data.base, r)
+    }
```
survived: Why did you choose a new way of defining commit function? In all other proofs, commit function takes public data and source of randomness and produces the commitment

generic-ec-zkp/src/dlog_eq.rs
```diff
@@ -0,0 +1,277 @@
...
+/// [`dlog_eq`](crate::dlog_eq)
+///
+/// Based on "Zero-Knowledge Proofs Notes", 2024 by Jorge L. Villar
+pub mod interactive {
+    use generic_ec::{Curve, Point, Scalar, SecretScalar};
+
+    /// Commitment to `H`, sent in the first round
+    pub type Commitment<E> = (Point<E>, Point<E>);
+    /// Challenge, sent to the Prover in the first round
+    pub type Challenge<E> = Scalar<E>;
```
survived, line 136: There's no function to generate a challenge. We usually add this function, even if all it does is sample a scalar

generic-ec-zkp/src/dlog_eq.rs
```diff
@@ -0,0 +1,277 @@
...
+        let com2 = data.base * proof.res - data.exp * proof.ch;
+
+        let seed = udigest::inline_struct!(TAG {
+            shared_state,
+            data,
+            com1,
+            com2,
+        });
+        let mut rng = rand_hash::HashRng::<D, _>::from_seed(seed);
+        let ch = Scalar::random(&mut rng);
```
survived: You probably want to use [`Scalar::from_hash`](https://docs.rs/generic-ec/latest/generic_ec/struct.Scalar.html#method.from_hash) instead

generic-ec-zkp/src/dlog_eq.rs
maurges: 1. Because I was mirroring how it's done in the tBLS paper, where `r` is generated externally.
2. Because otherwise we would have to make an explicit private commitment structure. Actually this might be a good thing
3. I felt like having non_interactive::commit be independent of rng dependency is a nice goal

I don't have a strong opinion on any of this though, just giving my thought process

generic-ec-zkp/src/dlog_eq.rs
maurges: You're right, I missed that part. Though I wonder if this generality is worth the worse ergonomics in the common case where g is just the habitual generator. I just had an idea how to make a smart constructor to return the ergonomics back..

generic-ec-zkp/src/dlog_eq.rs
maurges: Fixed

generic-ec-zkp/src/dlog_eq.rs
survived: I think, ideally we should make the API hard to misuse, in particular, exposed functions should contain most of the workflow. If we expect caller to do something correctly and provide it as inputs, that's where something can go wrong, so it's easier to misuse

generic-ec-zkp/src/dlog_eq.rs
survived: Yeah the constructor may work

generic-ec-zkp/src/dlog_eq.rs
```diff
@@ -98,16 +92,33 @@
...
-    pub exp: Point<E>,
+    /// `G`, the base for one logarithm
+    pub base1: Point<E>,
+    /// `X`, the value inside one logarithm
+    pub exp1: Point<E>,
+    /// `H`, the base for the other logarithm
+    pub base2: Point<E>,
+    /// `Z`, the value inside the other logarithm
+    pub exp2: Point<E>,
+}
```
survived: Whyyy do you use exp notation here?? We are multiplicative adepts, no?

generic-ec-zkp/src/dlog_eq.rs
```diff
@@ -98,16 +92,33 @@
...
+
+impl<E: Curve> Data<E> {
+    /// Create the data for the common case where `G` is the principal group
+    /// generator, and `x` is a secret key
+    ///
+    /// In this case, we set `H = base` as base2 and `Z = G * x` as exp2
+    pub fn from_secret_key(x: &generic_ec::SecretScalar<E>, base: Point<E>) -> Data<E> {
+        Self {
+            base1: Point::generator().into(),
+            exp1: Point::generator() * x,
```
survived: Here you don't write `Point::generator().pow(x)`...

generic-ec-zkp/src/dlog_eq.rs
maurges: Yeah I find it weird too (= But actually, what's the appropriate terminology for the additive notation? I've only ever seen multiplicative and logarithms

generic-ec-zkp/src/dlog_eq.rs
maurges: Yeah you're actually right, the only thing you can do with those nonces is generate and use them in a proof once, so they should be rng. Fixed

generic-ec-zkp/src/dlog_eq.rs
maurges: Did the constructor in an earlier commit, have a look

generic-ec-zkp/src/dlog_eq.rs
survived: I think you can call it `point1` and `point2`. Calling it `mult` would be misleading

generic-ec-zkp/src/dlog_eq.rs
```diff
@@ -0,0 +1,305 @@
...
+    pub use super::InvalidProof;
+
+    /// First round of the protocol: commit to `H` by itself. Produces
+    /// [`Commitment`] to be sent
+    ///
+    /// `rng` is used to generate the nonce for the private commitment
+    pub fn commit<E: Curve>(
+        base1: Point<E>,
+        base2: Point<E>,
+        rng: &mut impl rand_core::RngCore,
```
survived: Should we require that `rng` is `CryptoRng`? It's important for ZK security that rng is reliable

survived: I've already found my piece that we don't have consistency in that, but just a note, that everywhere else in this library `rng` comes as the first argument, see [`Polynomial::{sample, sample_with_const_term}`](https://docs.rs/generic-ec-zkp/latest/generic_ec_zkp/polynomial/struct.Polynomial.html#method.sample)

generic-ec-zkp/src/dlog_eq.rs
```diff
@@ -0,0 +1,305 @@
...
+    /// Create a proof deterministically for given data
+    ///
+    /// - `shared_state` - shared data not known to this proof, used to protect
+    ///   from replay attacks
+    /// - `share` - `x`, secret value of discrete logarithm
+    /// - `data` - data to compute the proof for
+    /// - `rng` - used to generate a random nonce for proof
+    pub fn prove<D: digest::Digest, E: Curve>(
+        shared_state: &impl udigest::Digestable,
+        share: &SecretScalar<E>,
```
survived: Should it rather be just called `x`? `share` is specific to threshold-{DH, BLS} application, but here it might be not the case for generic application

generic-ec-zkp/src/dlog_eq.rs
```diff
@@ -0,0 +1,305 @@
...
+        data: Data<E>,
+        rng: &mut impl rand_core::RngCore,
+    ) -> Proof<E> {
+        let r = Scalar::random(rng);
+        let com1 = data.base1 * r;
+        let com2 = data.base2 * r;
```
survived: You could replace it with calling `interactive::commit_data`

survived: Ah, I just realized that non_interactive proof is not just a naive compilation of interactive functions... Then it makes sense to not reuse `commit_data`

generic-ec-zkp/src/dlog_eq.rs
```diff
@@ -0,0 +1,305 @@
...
+#[cfg_attr(feature = "udigest", derive(udigest::Digestable), udigest(bound = ""))]
+pub struct Data<E: Curve> {
+    /// `G`, the base for one logarithm
+    pub base1: Point<E>,
+    /// `X`, the value inside one logarithm
+    pub exp1: Point<E>,
+    /// `H`, the base for the other logarithm
+    pub base2: Point<E>,
+    /// `Z`, the value inside the other logarithm
+    pub exp2: Point<E>,
```
survived: I think it's called `y` in the paper, not `z`?

survived: I see that you use it everywhere, so I leave using this letter up to you :camera_flash: :raised_eyebrow:

tests/tests/dlog_eq.rs
```diff
@@ -0,0 +1,91 @@
...
+    fn passing_test<E: generic_ec::Curve>() {
+        let mut rng = rand_dev::DevRng::new();
+
+        let secret_share = generic_ec::SecretScalar::<E>::random(&mut rng);
+
+        let base = generic_ec::Point::generator() * generic_ec::Scalar::random(&mut rng);
+        let data = dlog_eq::Data::from_secret_key(&secret_share, base);
+
+        let (comm, pcomm) = dlog_eq::commit_data(data, &mut rng);
+        let challenge = generic_ec::Scalar::random(&mut rng);
```
survived: can you instead use challenge function provided in `generic_ec_zkp::dlog_eq::interactive` module?

tests/tests/dlog_eq.rs
```diff
@@ -0,0 +1,91 @@
...
+        let mut rng = rand_dev::DevRng::new();
+        let shared_state = "shared state";
+
+        let secret_share = generic_ec::SecretScalar::random(&mut rng);
+
+        let base = generic_ec::Point::generator() * generic_ec::Scalar::random(&mut rng);
+        let data = dlog_eq::Data::from_secret_key(&secret_share, base);
+
+        let proof = dlog_eq::prove::<D, E>(&shared_state, &secret_share, data, &mut rng);
+        dlog_eq::verify::<D, E>(&shared_state, data, proof).unwrap();
```
survived: Usually `E` generic param goes first, and then `D` follows, like as you wrote in test function:

https://github.com/LFDT-Lockness/generic-ec/blob/7fec99d254c1e06220e0cb8aa859ff8f32158466/tests/tests/dlog_eq.rs#L53-L54

Maybe change order in `non_interactive` functions? It's a good idea to keep that order consistent

generic-ec-zkp/src/dlog_eq.rs
maurges: I forgot to change some terminology from bls, keep a look out for it!

generic-ec-zkp/src/dlog_eq.rs
maurges: Yes, I even left a comment about it at the top of the non-interactive module

generic-ec-zkp/src/dlog_eq.rs
maurges: Actually both Villar and tBLS call them y and z, so the `z` part is correct, it's me using `X` that is wrong (= But I wrote it this way to have our usual notation of `X = G * x`

tests/tests/dlog_eq.rs
maurges: On the other hand, we usually keep a parameter that is replaced with `_` last, like [here](https://docs.rs/cggmp21/latest/cggmp21/signing/struct.SigningBuilder.html#method.set_derivation_path_with_algo). I don't care either way for this function though

generic-ec-zkp/src/dlog_eq.rs
maurges: Hm, I agree about "mult" or even "product". "point" is a bit too generic, especially since everything here is a point.

And what about "base"? It's base in the sense of a base being exponentiated. I could rename them to "gen1" and "gen2" maybe, your thoughts?

generic-ec-zkp/src/dlog_eq.rs
maurges: Damn, we do here. We use it as the last argument in cggmp and ku though, argh. And in the whole of paillier-zk

generic-ec-zkp/src/dlog_eq.rs
survived: The paper uses x/y here
 
image

generic-ec-zkp/src/dlog_eq.rs
survived: In cggmp21, rng typically comes first 
image

survived: I guess all the code I write, rng goes first, and all code you write, rng goes last. So no wonder it's the case in ku23/paillier-zk

generic-ec-zkp/src/dlog_eq.rs
maurges: Ah, indeed. It's a lowercase `x` though. ;-) Let's just consider it a mix of those two.

generic-ec-zkp/src/dlog_eq.rs
survived: Also I believe your habit of bringing rng to the tail of arguments is your habit from haskell :stuck_out_tongue: In Rust, I believe it's typical to put rng in the beginning

generic-ec-zkp/src/dlog_eq.rs
survived: product should be good!

or prod1/prod2

survived: Hmm, that's true, I didn't think about `base`, it indeed comes from evil exp notation. I like your idea, we should use gen1/gen2 instead!

survived (approved):

PR #55, opened 2025-01-29
Constant-time scalar sampling & hash-to-scalar

survived:
We currently rely on actual curve implementation (i.e. rust crypto or dalek repos) to provide a function for random scalar generation from source of randomness.

However, in the code, we implicitly rely that random scalar generation is reproducible: we expect that when a reproducible PRNG is provided (such as HashRng) with the same seed, then output scalar is the same on all platforms. It's not guaranteed at all: a library may actually run platform-dependent algorithm for scalar generation, for instance, depending on whether it's x64 or x86 platform.

It is crucial to provide a reproducible `Scalar::random` as it is used in `Scalar::from_hash`, which in return is used in all non-interactive ZK proofs to derive a challenge.

This PR addresses this by implementing random scalar generation within the library. We simply take a sufficiently large bytestring, and reduce it modulo curve prime (sub)group order, in compliance with [RFC9380](https://www.rfc-editor.org/rfc/rfc9380.html#name-hashing-to-a-finite-field).

As a side effect, sampling a scalar is constant-time (which was not the case before)

### DIFF
--- a/generic-ec-curves/src/rust_crypto/scalar.rs
+++ b/generic-ec-curves/src/rust_crypto/scalar.rs
@@ -86,6 +86,10 @@ impl<E: CurveArithmetic> One for RustCryptoScalar<E> {
 
 #[cfg(feature = "secp256k1")]
 impl FromUniformBytes for RustCryptoScalar<k256::Secp256k1> {
+    /// 48 bytes
+    ///
+    /// `L = ceil((ceil(log2(q)) + k) / 8) = ceil((256 + 128) / 8) = 48` bytes are enough to
+    /// guarantee the uniform distribution
     type Bytes = [u8; 48];
     fn from_uniform_bytes(bytes: &Self::Bytes) -> Self {
         let mut bytes_be = [0u8; 64];
@@ -95,6 +99,10 @@ impl FromUniformBytes for RustCryptoScalar<k256::Secp256k1> {
 }
 #[cfg(feature = "secp256r1")]
 impl FromUniformBytes for RustCryptoScalar<p256::NistP256> {
+    /// 48 bytes
+    ///
+    /// `L = ceil((ceil(log2(q)) + k) / 8) = ceil((256 + 128) / 8) = 48` bytes are enough to
+    /// guarantee the uniform distribution
     type Bytes = [u8; 48];
     fn from_uniform_bytes(bytes: &Self::Bytes) -> Self {
         BytesModOrder::from_be_bytes_mod_order(bytes)
@@ -102,6 +110,10 @@ impl FromUniformBytes for RustCryptoScalar<p256::NistP256> {
 }
 #[cfg(feature = "stark")]
 impl FromUniformBytes for RustCryptoScalar<stark_curve::StarkCurve> {
+    /// 48 bytes
+    ///
+    /// `L = ceil((ceil(log2(q)) + k) / 8) = ceil((256 + 128) / 8) = 48` bytes are enough to
+    /// guarantee the uniform distribution
     type Bytes = [u8; 48];
     fn from_uniform_bytes(bytes: &Self::Bytes) -> Self {
         BytesModOrder::from_be_bytes_mod_order(bytes)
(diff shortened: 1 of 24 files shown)


generic-ec-curves/src/rust_crypto/scalar.rs
```diff
@@ -84,17 +84,31 @@ impl<E: CurveArithmetic> One for RustCryptoScalar<E> {
...
-        loop {
-            rng.fill_bytes(bytes.as_mut());
-
-            if let Some(scalar) = <E::Scalar as PrimeField>::from_repr_vartime(bytes.clone()) {
-                break Self(scalar);
-            }
-        }
+#[cfg(feature = "secp256k1")]
+impl FromUniformBytes for RustCryptoScalar<k256::Secp256k1> {
+    type Bytes = [u8; 48];
```
maurges: Maybe add a comment calculating the size like for ed25519?

maurges:
Is it maybe worth to keep old random function as fast_random?

survived:
Why do you assume it’s faster?

maurges:
> Why do you assume it’s faster?

Because I associate constant-time with slowness (=

I assume the authors of the underlying libraries optimized them more than we did. Also they might have other upsides, like consuming less entropy and having better distribution. The only one I care about from these is speed, for tests

But I don't care about it much. If you think it's not worth it to complicate the api, I don't mind

survived:
In the tests, I think we want to use a reproducible function, to have reproducible tests

survived:
It's interesting to see if there's any significant performance difference, I will do the experiments

survived:
Did benchmarks in 4639ab1, results:

```text
random/secp256k1/v04    time:   [34.735 ns 34.743 ns 34.753 ns]
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) high mild
  3 (3.00%) high severe
random/secp256k1/latest time:   [69.344 ns 69.417 ns 69.506 ns]
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  2 (2.00%) high severe

random/secp256r1/v04    time:   [34.955 ns 35.007 ns 35.113 ns]
Found 12 outliers among 100 measurements (12.00%)
  3 (3.00%) low mild
  3 (3.00%) high mild
  6 (6.00%) high severe
random/secp256r1/latest time:   [131.63 ns 131.69 ns 131.76 ns]
Found 13 outliers among 100 measurements (13.00%)
  5 (5.00%) high mild
  8 (8.00%) high severe

random/stark/v04        time:   [1.8662 µs 1.8687 µs 1.8711 µs]
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low severe
  4 (4.00%) low mild
  1 (1.00%) high mild
random/stark/latest     time:   [6.4075 µs 6.4093 µs 6.4114 µs]
Found 16 outliers among 100 measurements (16.00%)
  5 (5.00%) high mild
  11 (11.00%) high severe

random/ed25519/v04      time:   [163.51 ns 163.55 ns 163.60 ns]
Found 9 outliers among 100 measurements (9.00%)
  2 (2.00%) low mild
  3 (3.00%) high mild
  4 (4.00%) high severe
random/ed25519/latest   time:   [155.30 ns 155.35 ns 155.41 ns]
Found 12 outliers among 100 measurements (12.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  9 (9.00%) high severe
```

Yeah it became 2-6 times worse, except ed25519, which became faster 😄 Given that the difference is significant, it does make sense to provide vartime random generation

survived:
Ah, ed25519 became faster because they [use the same method](https://docs.rs/curve25519-dalek/4.1.3/src/curve25519_dalek/scalar.rs.html#597-601) for random scalar generation, except that they reduce 64 bytes instead of 48 bytes

survived:
@maurges added vartime random method (I don't think we can use it in the tests though)

maurges:
>  (I don't think we can use it in the tests though)

Because of reproducibility?

maurges (approved):

survived:
> Because of reproducibility?

Yes exactly, in most cases we want tests to be reproducible

PR #57, opened 2026-03-10
Fix hardcoded scalar size in Straus NAF computation

sridhar-panigrahi:
The `add_scalar` method in `NafMatrix` hardcodes `x_u64[0..4]` when reading scalar bytes, which only works for 32-byte scalars. This panics for curves with different scalar sizes like P-384 or P-521.

Use the actual scalar byte length instead.

### DIFF
--- a/generic-ec-curves/src/lib.rs
+++ b/generic-ec-curves/src/lib.rs
@@ -6,7 +6,7 @@
 //! [`generic-ec` crate]: https://docs.rs/generic-ec
 
 #![cfg_attr(not(test), deny(clippy::unwrap_used, clippy::expect_used))]
-#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
 #![forbid(missing_docs)]
 #![no_std]
 
--- a/generic-ec/src/lib.rs
+++ b/generic-ec/src/lib.rs
@@ -184,7 +184,7 @@
 #![cfg_attr(not(test), forbid(unused_crate_dependencies))]
 #![cfg_attr(not(test), deny(clippy::unwrap_used, clippy::expect_used))]
 #![no_std]
-#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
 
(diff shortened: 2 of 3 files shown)


sridhar-panigrahi:
@survived , please let me know your thoughts on this !

survived:
Hi @sridhar-panigrahi. What you propose looks like a reasonable change, but we cannot test it since there's no curve with scalar_len != 32

Bringing curves with larger scalar sizes is something we would like to do, but it's a broader topic than just updating NAF computations: we need to add an actual curve to the crate, update all the tests, see what gets broken (probably something else besides multiscalar multiplication), and otherwise carefully check all code to make sure it makes no assumptions on scalar size.

sridhar-panigrahi:
Hi @survived, thanks for the detailed feedback!

I completely understand — supporting larger scalar sizes is a broader effort that goes well beyond just this NAF fix. I did a quick scan and noticed there are similar assumptions in other places too (e.g., `FromUniformBytes` hardcoding 48-byte buffers based on 256-bit curves, the `scalar_from_*_bytes_mod_order_reducing_*` utils in `generic-ec-curves`, etc.).

Would you be open to merging this as a small incremental fix for now? It doesn't break anything for the current 32-byte curves, and it removes one assumption that would otherwise panic for larger scalars.

I'd also be happy to help with the broader work — adding a curve with a larger scalar size, auditing the rest of the codebase for similar hardcoded assumptions, and updating the tests. If you'd prefer that as a single comprehensive PR instead, I can work on that too. Just let me know what approach you'd prefer!

survived:
I would prefer to defer merging this PR until we start adding curves with larger scalar sizes. I want the merged code to be tested against various scalar sizes to make sure everything works as expected. Given that PR concerns one line, we should be able to merge it without merge conflicts once we start working on it.

survived:
I added #58 issue to track this project

sridhar-panigrahi:
Thanks @survived, that makes total sense — deferring until it can be properly tested is the right call.

I'll keep an eye on #58. Happy to pick up work on it when you're ready to start adding larger scalar curves — whether that's adding the curve implementation, auditing other hardcoded assumptions, or updating tests. Just let me know!

survived:
Done in #59

Issue #58, opened 2026-03-12
Add support of curves with larger scalar sizes


survived:
Currently, the crate only supports curves with scalars of 32 bytes. We want to expand it to work with curves that have scalar size $\ge 32$ bytes, e.g. <s>curve448,</s> ed448, P-384, P-521, and so on.

The size of the scalar has been hardcoded into library in several places, therefore if we simply add a curve with larger scalar size, library will break in some places causing runtime panics.

Expected development path:
1. Add curves with different scalar sizes to the library, update all tests to use these curves, and observe what tests get broken
2. Fix the issues until tests pass
3. Update [`cggmp24`](https://github.com/LFDT-Lockness/cggmp21) and [`givre`](https://github.com/lfdt-lockness/givre) crates to support these curves, update all tests to use these curves and make sure all tests pass
4. Carefully review the library to identify any other places where we could make assumptions on the scalar size

Once all of this is done, we can release a new version of `generic-ec[-curves]` with the added curves. Preferably, that would be a release without any breaking changes in API unless we absolutely have to break the API.

Related work:
- Fix hardcoded scalar size in multiscalar multiplication #57

If you're willing to take part in development, reach out to us in discord [`#lockness-contribute` channel](https://discord.com/channels/905194001349627914/1285268686147424388).

sridhar-panigrahi:
Hey @survived ,  I've started working on this! Opened #59 which adds secp384r1 (P-384)
   as the first curve with a larger scalar size (48 bytes). All existing
  tests are instantiated for the new curve and pass — no breakage found
  beyond the Straus NAF fix already in #57.

  Happy to continue with the next steps from the development path — adding
  more curves with different scalar sizes and then updating cggmp24 and
  givre.

survived:
Hi @sridhar-panigrahi, thanks for your help in this issue! I've released `generic-ec` with the secp384r1 curve and NAF changes you made.

What concerns this PR, the library could still benefit from other "big" curves such as curve448 (so we can integrate it into `givre`), and P-521. I leave issue open for anyone to pick it.

sridhar-panigrahi:
@survived That's awesome, thanks for getting it released\! I'm definitely interested in continuing with this — curve448 sounds like a good next step since it'd help with `givre`. I'll dig into it and open a PR when it's ready. P-521 can follow after that.

sridhar-panigrahi:
Hey @survived, just opened #61 adding Curve448 support! Used
[...]

PR #59, opened 2026-03-12
Add secp384r1 (NIST P-384) curve support

sridhar-panigrahi:
## Summary

Hey! Following up on the discussion in #lockness-contribute and the plan laid out in #58 — this PR adds secp384r1 (NIST P-384) as the first curve with a larger scalar size (48 bytes / 384 bits) to the library.

The goal here was to actually plug in a bigger curve, run the full test suite against it, and see what breaks. The good news is: nothing broke. The existing scalar pipeline and the Straus NAF fix from #57 handle the 48-byte scalars cleanly.

### What's in here

- **p384 crate wired in** — new `secp384r1` feature flag through both `generic-ec-curves` and `generic-ec`, following the same RustCrypto pattern as the existing curves
- **Scalar traits implemented** — `CurveName`, `FromUniformBytes` (L=64 bytes per RFC 9380), `BytesModOrder`, `Reduce<48>` via `U384`, and `NoInvalidPoints`
- **New reduction utilities** — `scalar_from_{be,le}_bytes_mod_order_reducing_48` for Horner-method mod-order reduction using 48-byte chunks (same pattern as the `_reducing_32` variants)
- **Every test instantiated for secp384r1** — scalar ops, point ops, serde round-trips, Straus multiscalar multiplication, affine coordinates, dlog_eq ZK proofs, `Reduce<48>`, and `from_bytes_mod_order` with various input sizes
- **Benchmarks added** for the new curve

### Test results

All 180 tests pass, including all the new secp384r1 ones. Clippy is clean too. No hardcoded scalar-size assumptions surfaced beyond the one already fixed in #57.

### What's next (per the plan in #58)

- Look at adding more curves with different scalar sizes (curve448, P-521) to stress-test further
- Update `cggmp24` and `givre` to use these curves and make sure their tests pass
- Do a careful review for any remaining scalar-size assumptions that the test suite might not cover

Happy to adjust anything based on your feedback!

Relates to #58
Depends on #57

## Test plan

- [x] `cargo test --all-features` — 180/180 tests pass
- [x] `cargo clippy --all-features` — no warnings
- [x] Scalar byte round-trips (BE/LE) work for 48-byte scalars
- [x] `from_bytes_mod_order` works with various input lengths (8, 16, 32, 48, 64, 80, 96, 128, 256, 512 bytes and random lengths)
- [x] Straus multiscalar multiplication works with 384-bit NAFs
- [x] Serde serialization/deserialization round-trips
- [x] Affine coordinate access (x, y, parity) works
- [x] dlog_eq zero-knowledge proofs pass (both interactive and non-interactive)

### DIFF
--- a/generic-ec-curves/src/rust_crypto/mod.rs
+++ b/generic-ec-curves/src/rust_crypto/mod.rs
@@ -20,7 +20,7 @@ use generic_ec_core::{
 use subtle::{ConditionallySelectable, ConstantTimeEq};
 use zeroize::{DefaultIsZeroes, Zeroize};
 
-#[cfg(any(feature = "secp256k1", feature = "secp256r1", feature = "stark"))]
+#[cfg(any(feature = "secp256k1", feature = "secp256r1", feature = "secp384r1", feature = "stark"))]
 use sha2::Sha256;
 
 pub use self::{curve_name::CurveName, point::RustCryptoPoint, scalar::RustCryptoScalar};
@@ -66,6 +66,12 @@ pub type Secp256k1 = RustCryptoCurve<k256::Secp256k1, ExpandMsgXmd<Sha256>>;
 /// Based on [p256] crate
 #[cfg(feature = "secp256r1")]
 pub type Secp256r1 = RustCryptoCurve<p256::NistP256, ExpandMsgXmd<Sha256>>;
+
+/// secp384r1 curve (NIST P-384)
+///
+/// Based on [p384] crate
+#[cfg(feature = "secp384r1")]
+pub type Secp384r1 = RustCryptoCurve<p384::NistP384, ExpandMsgXmd<sha2::Sha384>>;
 
 /// Stark curve
 ///
@@ -175,6 +181,13 @@ unsafe impl NoInvalidPoints for Secp256r1 {}
 /// Safe because:
 /// - RustCrypto curves are always on curve:
 ///   generic-ec-curves/src/rust_crypto/point.rs:60
+/// - p384 is prime order and so is always torsion-free:
+///   <https://github.com/RustCrypto/elliptic-curves/blob/7a71e403e49fbe92d6b2ae8fe3eabbfdef124975/primeorder/src/projective.rs#L172-L174>
+#[cfg(feature = "secp384r1")]
+unsafe impl NoInvalidPoints for Secp384r1 {}
+/// Safe because:
+/// - RustCrypto curves are always on curve:
+///   generic-ec-curves/src/rust_crypto/point.rs:60
 /// - stark is prime order and so is always torsion-free:
 ///   <https://github.com/RustCrypto/elliptic-curves/blob/7a71e403e49fbe92d6b2ae8fe3eabbfdef124975/primeorder/src/projective.rs#L172-L174>
 #[cfg(feature = "stark")]
@@ -187,7 +200,7 @@ mod tests {
         Curve,
     };
 
-    use super::{Secp256k1, Secp256r1, Stark};
+    use super::{Secp256k1, Secp256r1, Secp384r1, Stark};
 
     /// Asserts that `E` implements `Curve`
     fn _impls_curve<E: Curve>() {}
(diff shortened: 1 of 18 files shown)


sridhar-panigrahi:
@survived , let me know your thoughts on this !

generic-ec-curves/src/rust_crypto/mod.rs
```diff
@@ -20,7 +20,7 @@ use generic_ec_core::{
 use subtle::{ConditionallySelectable, ConstantTimeEq};
 use zeroize::{DefaultIsZeroes, Zeroize};
 
-#[cfg(any(feature = "secp256k1", feature = "secp256r1", feature = "stark"))]
+#[cfg(any(feature = "secp256k1", feature = "secp256r1", feature = "secp384r1", feature = "stark"))]
 use sha2::Sha256;
```
survived, line 24: I don't see any place where `Sha256` is used for secp384r1. You use `sha2::Sha384` instead.

survived (commented):
Hi @sridhar-panigrahi, the PR looks good on my end. There's one unnecessary change I found, but rather than that it's good! 

Could you do GPG-signing on your commits? We can't merge unsigned commits unfortunately (it's mentioned in [CONTRIBUTING.md](https://github.com/LFDT-Lockness/generic-ec/blob/m/CONTRIBUTING.md)).

In meantime, I'll update cggmp to use this curve and see if all works and it's able to generate valid P-384 ECDSA signatures.

generic-ec-curves/src/utils.rs
```diff
@@ -130,6 +130,114 @@ where
...
+/// Interprets `bytes` as little-endian encoding of an integer, takes it modulo curve (prime)
+/// order and returns scalar `S`
+///
+/// Works with scalars for which only [`Reduce<48>`][Reduce] is defined.
+///
+/// Takes:
+/// * Little-endian `bytes` representation of the integer
+/// * Scalar `one = 1`
+pub fn scalar_from_le_bytes_mod_order_reducing_48<S>(bytes: &[u8], one: &S) -> S
+where
```
maurges: This basically copies the code for `_32` function. Should we make it one function generic over `N`?

sridhar-panigrahi:
Hi @survived and @maurges, thanks for the feedback! I've gone ahead and addressed both of the review comments:

- Removed `secp384r1` from the `Sha256` cfg import guard since P-384 uses `sha2::Sha384` instead (thanks for catching that @survived!)
- Unified the duplicated `_32` and `_48` helpers into a single generic `scalar_from_{le,be}_bytes_mod_order_reducing<S, const N: usize>` to avoid the code duplication (great suggestion @maurges!)

All commits are GPG-signed as well. Let me know if there's anything else you'd like me to change!

maurges:
Thanks for the change. It seems you haven't signed off on the latest commit, that's why the CI is failing now

sridhar-panigrahi:
@maurges Added the DCO sign-off to the latest commit. CI checks are passing now.

sridhar-panigrahi:
@survived @maurges  , please do let me know if you think there could be any changes
.

maurges:
The code looks good, I resolved the remaining comments. The CI is still failing - it seems you updated the docs in lib.rs without updating the readme accordingly. Currently they need to manually be kept in sync, and you can test that on your machine by running `cd generic-ec; cargo rdme -r ../README.md --check` (the same command as in CI)

survived:
> by running `cd generic-ec; cargo rdme -r ../README.md --check` (the same command as in CI)

We have `make readme` for that!

sridhar-panigrahi:
@maurges @survived Thanks for the heads up! I've synced the README with 
  cargo rdme and pushed the fix. Good to know about make readme — I'll use 
  that going forward. CI is green now.

survived (approved):
All looks good! I tested that cggmp can successfully yield a valid signature on secp384r1 curve in https://github.com/LFDT-Lockness/cggmp21/pull/173

sridhar-panigrahi:
@survived Awesome, thanks for testing and approving! Great to hear the  
  P-384 ECDSA signatures work end-to-end with cggmp.

maurges (approved):

PR #61, opened 2026-03-23
Add Ed448 (Goldilocks) curve support

veyron-kairo:
Following up on #58 — updated based on reviewer feedback to use **ed448** instead of curve448.

## What changed

- Renamed to `Ed448`, `CURVE_NAME = "ed448"`, feature flag `curve-ed448`
- Scalar encoding is now 57 bytes (RFC 8032) instead of 56
- `Zeroize` derived on the curve struct, `Point`, and `Scalar`
- `FromUniformBytes` now uses 84 bytes (k = 224 per RFC 9380)
- `Reduce<57>` and `Reduce<114>` replacing the old `Reduce<56>`/`Reduce<112>`

All tests pass.


sridhar-panigrahi:
@survived  , please let me know you thoughts on this !

survived:
Hey @sridhar-panigrahi, sorry, it will take me some time to review the PR as I'm pressed by some other work. I should be able to get on this in a couple of weeks unless @maurges gets to it first!

sridhar-panigrahi:
it's ok @survived , I'll try to find and work on something even more valuable on this repo .

sridhar-panigrahi:
@maurges , please let me know your thoughts on this !

generic-ec-curves/src/curve448.rs
```diff
@@ -0,0 +1,417 @@
...
+
+/// Pads a 112-byte LE scalar to 114 bytes (WideScalarBytes format)
+fn pad_to_114(bytes: &[u8; 112]) -> [u8; 114] {
+    let mut padded = [0u8; 114];
+    padded[..112].copy_from_slice(bytes);
+    padded
+}
+
+/// Curve448 (Goldilocks) curve
+#[derive(Clone, Copy, Debug, PartialEq, PartialOrd, Eq, Ord, Hash, Default)]
```
maurges: Why not derive `Zeroize` too?

generic-ec-curves/src/curve448.rs
```diff
@@ -0,0 +1,417 @@
...
+
+impl generic_ec_core::CompressedEncoding for Point {
+    type Bytes = [u8; 57];
+
+    fn to_bytes_compressed(&self) -> Self::Bytes {
+        use group::GroupEncoding;
+        let ga = self.0.to_bytes();
+        let mut bytes = [0u8; 57];
+        bytes.copy_from_slice(ga.as_ref());
+        bytes
```
maurges: GenericArray has an `Into<[u8; 57]>` impl that could be used here instead of explicit copying

generic-ec-curves/src/curve448.rs
```diff
@@ -0,0 +1,417 @@
...
+    fn zeroize(&mut self) {
+        // Replace with identity point
+        self.0 = group::Group::identity();
+    }
+}
+
+// --- Scalar ---
+
+/// Curve448 scalar
+#[derive(Default, Clone, Copy, PartialEq, Eq)]
```
maurges: Why not derive `Zeroize` too?

generic-ec-curves/src/curve448.rs
```diff
@@ -0,0 +1,417 @@
...
+        subtle::ConstantTimeEq::ct_eq(&x.0, &ed448_goldilocks_plus::Scalar::ONE)
+    }
+}
+
+impl generic_ec_core::FromUniformBytes for Scalar {
+    /// 72 bytes
+    ///
+    /// `L = ceil((ceil(log2(q)) + k) / 8) = ceil((446 + 128) / 8) = 72` bytes are enough to
+    /// guarantee the uniform distribution (RFC 9380)
+    type Bytes = [u8; 72];
```
maurges: If I read the rfc right, the security level k here is 224, and thus L = 84

maurges:
I left a couple of comments, but the big question is whether we want this curve to be curve448 (56 byte encoded scalar) or ed448 (57 byte encoded scalar). I'm tending towards the second, since the biggest usage of this curve is for signatures after all, but maybe curve448 is more general and is easy enough to adapt to signatures too. Or maybe it's possible to have both

survived:
We def need ed448 not curve448, to align with existing ed25519

I still haven't had time to look at PR, but curve448 is in a twisted form in which there's no point addition defined. The only way around is by mapping points from curve448 to ed448, perform point addition, and map back. This is extremely inefficient, we should not do that.

survived:
Sorry, it's my bad, I wrote curve448 in #58. I meant ed448

maurges:
> The only way around is by mapping points from curve448 to ed448, perform point addition, and map back

If I understand right, here internally everything is treated as ed448 values, and is only encoded as curve448

sridhar-panigrahi:
Hey @survived and @maurges, I've updated the PR to address your feedback:

- Switched from curve448 to **ed448** as requested — renamed struct to `Ed448`, set `CURVE_NAME = "ed448"`, and changed scalar encoding to 57 bytes (RFC 8032 format)
- Added `zeroize::Zeroize` derive to the curve struct, `Point`, and `Scalar`
- Used `GenericArray::into()` in `to_bytes_compressed` instead of manual copy
- Fixed `FromUniformBytes` to use L = 84 bytes (k = 224, as you pointed out @maurges)
- Also renamed the feature flag from `curve448` to `ed448` (and `curve-curve448` to `curve-ed448` in the top-level crate)

Let me know if there's anything else to adjust!

generic-ec-curves/src/ed448.rs
```diff
@@ -0,0 +1,385 @@
...
+        ))
+    }
+}
+
+impl generic_ec_core::CompressedEncoding for Point {
+    type Bytes = [u8; 57];
+
+    fn to_bytes_compressed(&self) -> Self::Bytes {
+        use group::GroupEncoding;
+        self.0.to_bytes().into()
```
maurges, line 103: Nitpick: it's inconsistent that above you use a fully qualified call to `group::cofactor::CofactorGroup::is_torsion_free`, but here you import instead.

My personal preference is to use fully-qualified calls. But importing traits does make the implementation look more similar to ed25519, which makes this review easier. So use your judgement

generic-ec-curves/src/ed448.rs
```diff
@@ -0,0 +1,385 @@
...
+}
+
+impl core::fmt::Debug for Point {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        use generic_ec_core::CompressedEncoding;
+        f.debug_tuple("Point")
+            .field(&self.to_bytes_compressed())
+            .finish()
+    }
+}
```
maurges: What's the reasoning for an explicit Debug impl? Does it make it more similar to how the other curves in this crate behave? Or something else?

generic-ec-curves/src/ed448.rs
```diff
@@ -0,0 +1,385 @@
...
+    }
+    fn from_le_array_mod_order(bytes: &[u8; 57]) -> Self {
+        // `from_bytes_mod_order` only uses bytes[0..56] and ignores byte[56].
+        // Use `from_bytes_mod_order_wide` with zero-padding so all 57 bytes
+        // (= 456-bit LE integer) are correctly reduced mod the group order.
+        let mut wide = [0u8; 114];
+        wide[..57].copy_from_slice(bytes);
+        Self(ed448_goldilocks_plus::Scalar::from_bytes_mod_order_wide(
+            GenericArray::from_slice(&wide),
+        ))
```
maurges: You can create wide as `let mut wide = GenericArray::default()`, and then there's no need to convert

generic-ec-curves/src/ed448.rs
```diff
@@ -0,0 +1,385 @@
...
+
+    fn to_bytes_uncompressed(&self) -> Self::Bytes {
+        <Self as generic_ec_core::CompressedEncoding>::to_bytes_compressed(self)
+    }
+}
+
+impl generic_ec_core::Decode for Point {
+    fn decode(bytes: &[u8]) -> Option<Self> {
+        use group::GroupEncoding;
+        let ga = GenericArray::from_slice(bytes);
```
maurges, line 121: This will panic if the slice is too short. Better to use `try_from_slice` and return None

generic-ec-curves/src/ed448.rs
```diff
@@ -0,0 +1,385 @@
...
+    /// `L = ceil((ceil(log2(q)) + k) / 8) = ceil((446 + 224) / 8) = 84` bytes are enough to
+    /// guarantee the uniform distribution (RFC 9380)
+    type Bytes = [u8; 84];
+
+    fn from_uniform_bytes(bytes: &Self::Bytes) -> Self {
+        let mut bytes_le = [0u8; 114];
+        bytes_le[..84].copy_from_slice(bytes);
+        Self(ed448_goldilocks_plus::Scalar::from_bytes_mod_order_wide(
+            GenericArray::from_slice(&bytes_le),
+        ))
```
maurges: You can create a GenericArray directly instead of first writing to an array

generic-ec-curves/src/ed448.rs
```diff
@@ -0,0 +1,385 @@
...
+
+    fn from_be_bytes_exact(bytes: &Self::Bytes) -> Option<Self> {
+        let mut bytes = *bytes;
+        bytes.reverse();
+        Self::from_le_bytes_exact(&bytes)
+    }
+
+    fn from_le_bytes_exact(bytes: &Self::Bytes) -> Option<Self> {
+        Option::from(ed448_goldilocks_plus::Scalar::from_canonical_bytes(
+            GenericArray::from_slice(bytes),
```
maurges: This can panic in runtime if we ever mistake the lengths. But there is an instance `impl<'a, T, const N: usize> From<&'a [T; N]> for &'a GenericArray<T, ConstArrayLength<N>>`, so you should be able to simply write `bytes.into()` here

maurges (changes requested):
I marked three places with the usage of `GenericArray::from_slice`, but there are others, and all of them should be replaced

generic-ec-curves/src/ed448.rs
veyron-kairo: Fixed — added an explicit length check before calling `from_slice`, returning `None` on mismatch. (`generic-array 0.14` doesn't have `try_from_slice`, so the length check achieves the same panic-safe result.)

generic-ec-curves/src/ed448.rs
veyron-kairo: Fixed — now uses `GenericArray::default()` directly for all the wide-buffer cases.

generic-ec-curves/src/ed448.rs
veyron-kairo: Fixed — now creates `GenericArray::default()` and fills it with `copy_from_slice`.

generic-ec-curves/src/ed448.rs
veyron-kairo: Fixed for the `[u8; 57]` case using `bytes.into()`. For `[u8; 114]`, `generic-array 0.14` only implements `From<&[T; N]>` for specific sizes and 114 isn't among them, so used `GenericArray::default()` + `copy_from_slice` there instead.

generic-ec-curves/src/ed448.rs
veyron-kairo: Used the inherent `self.0.is_torsion_free()` method directly — avoids both the verbose fully-qualified call and any trait import.

generic-ec-curves/src/ed448.rs
veyron-kairo: Removed both the explicit `Debug` impls for `Point` and `Scalar` — looking at `ed25519.rs`, it has no custom `Debug` on its raw types either, so this keeps things consistent.

sridhar-panigrahi:
Thank you for the detailed review, @maurges! I've addressed all the feedback in the latest commit:

- Replaced all `GenericArray::from_slice` calls — `Decode::decode` now does an explicit length check and returns `None` on mismatch; fixed-size array inputs use `GenericArray::default()` + `copy_from_slice` or `bytes.into()` (for `[u8; 57]`, which has a `From` impl in `generic-array 0.14`; `[u8; 114]` does not, so the default+copy pattern is used there).
- Removed the explicit `Debug` impls for `Point` and `Scalar` — they were unnecessary and inconsistent with how `ed25519` handles it.
- Simplified `is_torsion_free` to call `self.0.is_torsion_free()` directly, which avoids both the verbose fully-qualified call and any extra trait import.

Let me know if anything else needs attention!

generic-ec-curves/src/ed448.rs
maurges: The old version is very unfortunate.

Instead of an explicit length check, I'd prefer a more « correct by construction » approach:

```
        let bytes: &[u8; 57] = bytes.try_into().ok()?;
        Option::from(ed448_goldilocks_plus::EdwardsPoint::from_bytes(bytes.into())).map(Self)
```

generic-ec-curves/src/ed448.rs
```diff
@@ -0,0 +1,374 @@
...
+        Self(ed448_goldilocks_plus::Scalar::from_bytes_mod_order_wide(
+            &wide,
+        ))
+    }
+    fn from_le_array_mod_order(bytes: &[u8; 114]) -> Self {
+        let mut wide = GenericArray::default();
+        wide.copy_from_slice(bytes);
+        Self(ed448_goldilocks_plus::Scalar::from_bytes_mod_order_wide(
+            &wide,
+        ))
```
maurges, line 372: This annoys me, that we can't get rid of the copy, because the From instance for 114 doesn't exist int this version of generic-array. But I don't see any solutions

maurges:
Looks good, only one small comment left. Also please avoid force-pushing unless absolutely necessary, as it makes it hard to review the changes incrementally with github

sridhar-panigrahi:
@maurges I was looking at this and had a thought — not sure if I'm missing something, so wanted to ask before suggesting anything.

Since `from_bytes_mod_order_wide` takes `&WideScalarBytes` (a reference, not an owned value), could `GenericArray::from_slice` work here? It converts `&[u8]` → `&GenericArray<u8, U114>` via a pointer cast with no copy, and the impl does exist in generic-array 0.14:

```rust
fn from_le_array_mod_order(bytes: &[u8; 114]) -> Self {
    Self(ed448_goldilocks_plus::Scalar::from_bytes_mod_order_wide(
        GenericArray::from_slice(bytes),
    ))
}
```

And similarly for `from_be_array_mod_order`, where the reversal copy is unavoidable but the second copy into `wide` might not be:

```rust
fn from_be_array_mod_order(bytes: &[u8; 114]) -> Self {
    let mut bytes = *bytes;
    bytes.reverse();
    Self(ed448_goldilocks_plus::Scalar::from_bytes_mod_order_wide(
        GenericArray::from_slice(&bytes),
    ))
}
```

I think the confusion might be that `From<[u8; 114]>` for an owned `GenericArray` doesn't exist in 0.14 (which is true), but since we only need a reference, `From<&[u8]>` for `&GenericArray` is enough. Could be wrong though — happy to hear if there's a reason this wouldn't work!

sridhar-panigrahi:
@maurges  , sorry for using force push 😅
from now onwards I'll make sure not to do force-push unless necessary .

PR #62, opened 2026-04-10
Add     
  hash-based commitment scheme to generic-ec-zkp

sridhar-panigrahi:
Hey Denis and  
  Maurges,
                                                                         
  This adds a hash_commitment module to generic-ec-zkp — a statistically 
  hiding, computationally binding commitment scheme generic over any
  Digest and Digestable message type.                                    
                                           
  The scheme computes H(domain_tag || nonce || msg) where the 32-byte    
  nonce is sampled fresh each time. Verification uses
  subtle::ConstantTimeEq throughout. It fits naturally alongside the     
  existing Schnorr PoK and dlog-equality proofs in the crate.

  The immediate motivation is threshold protocols like DKLs23, where each
   party needs to commit to its key-share contribution before the reveal
  round — without it a rushing adversary can bias the combined key. The  
  primitive is generic enough to be useful anywhere a commit-reveal round
   appears.

  Happy to adjust the API shape or feature gating based on your          
  preferences.

### DIFF
--- a/generic-ec-zkp/src/hash_commitment.rs
+++ b/generic-ec-zkp/src/hash_commitment.rs
@@ -1,0 +1,212 @@
+//! Hash-based Commitment Scheme $\mathcal{F}_\text{com}$
+//!
+//! A statistically hiding, computationally binding commitment scheme over arbitrary
+//! digestable values. The committer chooses a random nonce $r$ and publishes
+//! $\text{com} = H(\text{tag} \| r \| \text{msg})$; the opening reveals $r$ so
+//! any verifier can recompute and check the commitment.
+//!
+//! This primitive is used in threshold protocols such as DKLs23 key generation,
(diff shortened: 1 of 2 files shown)


sridhar-panigrahi:
@survived , @maurges , please let me know your thoughts on this !

maurges:
Hello @sridhar-panigrahi. I don't think this protocol currently has a place in generic-ec. Note that in [contribution guidelines](https://github.com/LFDT-Lockness/round-based/blob/m/CONTRIBUTING.md) we ask all contributors to contact us before doing any work. It's usually a good reflex to talk to the maintainers before engaging into any development, at least to make sure your views are aligned.

Currently we're unsure about the direction of generic-ec-zkp, what primitives should be included in it. Maybe in the future we will revisit the idea of this patch.

PR #63, opened 2026-04-18
feat(zkp): enforce NonZero scalar for dlog_eq generators

### LINKED ISSUE #56: generic-ec-zkp dlog_eq should use NonZero scalar as input
As someone using this api pointed out:

> Whoa, shouldn't we require in dlog_eq that generator is NonZero<Point<E>>? Otherwise the proof is shit...

rupeshca007:
Fixes LFDT-Lockness/generic-ec#56

Currently, the `dlog_eq` protocol accepts `Point<E>` for generators, which does not guarantee that the generator is non-zero at compile-time. If a zero point is somehow passed as a generator, the zero-knowledge proof becomes degenerate. 

This PR updates the `Data` struct, the `commit` functions, and the `from_secret_key` constructor to strictly require `NonZero<Point<E>>` as input. I have also updated the interactive and non-interactive tests to safely use `.unwrap()` when wrapping randomly generated points using `NonZero::from_point`.

### DIFF
--- a/tests/tests/dlog_eq.rs
+++ b/tests/tests/dlog_eq.rs
@@ -8,7 +8,7 @@ mod interactive {
 
         let secret_share = generic_ec::SecretScalar::<E>::random(&mut rng);
 
-        let base = generic_ec::Point::generator() * generic_ec::Scalar::random(&mut rng);
+        let base = generic_ec::NonZero::from_point(generic_ec::Point::generator() * generic_ec::Scalar::random(&mut rng)).unwrap();
         let data = dlog_eq::Data::from_secret_key(&secret_share, base);
 
         let (comm, pcomm) = dlog_eq::commit_data(&mut rng, data);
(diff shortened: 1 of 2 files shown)


tests/tests/dlog_eq.rs
```diff
@@ -8,7 +8,7 @@ mod interactive {
 
         let secret_share = generic_ec::SecretScalar::<E>::random(&mut rng);
 
-        let base = generic_ec::Point::generator() * generic_ec::Scalar::random(&mut rng);
+        let base = generic_ec::NonZero::from_point(generic_ec::Point::generator() * generic_ec::Scalar::random(&mut rng)).unwrap();
```
maurges, line 11: Here and below, you could instead use `let base = generic_ec::Point::generator() * generic_ec::NonZero::<generic_ec::Scalar<E>>::random(&mut rng);` to get a non-zero point

maurges:
Thanks for this patch! The changes mostly look good, but since it is a breaking change in the API, it will take us some more time to figure out how to merge and release them

PR #71, opened 2026-05-29
Pin all workflows

maurges:


.github/workflows/rust.yml
```diff
@@ -144,15 +144,15 @@ jobs:
         - all-features
         - only-explicit-features
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 #v3
     - name: Check semver
-      uses: obi1kenobi/cargo-semver-checks-action@v2
+    - uses: obi1kenobi/cargo-semver-checks-action@6b69fcf40e9b5fb17adeb57e4b6ecd020649a239 #v2
```
survived: a typo here, added a `-` that is not needed

maurges: Yep, I was fixing the script to prevent that

survived (approved):

PR #90, opened 2026-06-11
Add Zeroize impl for encoded point and scalar

maurges:
These values are sometimes used in secure contexts like KEM, and it would be useful to wrap then in Zeroizing

### DIFF
--- a/generic-ec/src/encoded.rs
+++ b/generic-ec/src/encoded.rs
@@ -1,8 +1,11 @@
 use core::{fmt, ops};
 
+use zeroize::Zeroize;
+
 use crate::{as_raw::AsRaw, core::ByteArray, Curve};
 
 /// Bytes representation of an elliptic point
+#[derive(Zeroize)]
 pub struct EncodedPoint<E: Curve>(EncodedPointInner<E>);
 
 impl<E: Curve> EncodedPoint<E> {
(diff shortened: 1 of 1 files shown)


generic-ec/src/encoded.rs
```diff
@@ -67,6 +70,24 @@ impl<E: Curve> AsRef<[u8]> for EncodedPoint<E> {
...
+}
+
+impl<E: Curve> Zeroize for EncodedPointInner<E> {
+    fn zeroize(&mut self) {
+        match self {
+            EncodedPointInner::Compressed(a) => a.as_mut().zeroize(),
+            EncodedPointInner::Uncompressed(a) => a.as_mut().zeroize(),
+        }
+    }
+}
```
survived, line 89: Does it ever happen that an elliptic point represents a sensitive info? I don't recall any such case

generic-ec/src/encoded.rs
```diff
@@ -133,3 +154,9 @@ impl<E: Curve> AsRaw for EncodedScalar<E> {
         &self.0
     }
 }
+
+impl<E: Curve> Zeroize for EncodedScalar<E> {
+    fn zeroize(&mut self) {
+        self.as_mut().zeroize()
+    }
+}
```
survived, line 162: we might wanna add `SecretScalar::{to_be_bytes, to_le_bytes}` methods that return `Zeroizing<Box<EncodedScalar>>` (Box cuz otherwise Rust will leave copies of encoded scalar everywhere on the stack)

generic-ec/src/encoded.rs
survived: In this case, we would need a `SecretPoint<E>`...

maurges: Yes, like in ECIES or ECDH when a point is fed to KDF - that's a secret info that we want to zeroize

generic-ec/src/encoded.rs
maurges: Interesting idea. Right now I'm thinking about a SecretPoint and it would make a better and safer interface in this case

generic-ec/src/encoded.rs
survived: Right... But I see a little value in adding Zeroize to EncodedPoint, when the point itself is not Zeroize... But making a point Zeroize brings little value without adding SecretPoint...

survived: Ah the point is Zeroize already! Ok, then let's do `Zeroize for EncodedPoint` and leave `SecretPoint` as a thought for the future

PR #92, opened 2026-06-17
Add SecretPoint. Improve SecretScalar

maurges:
- SecretPoint is like SecretScalar but for points
- Add Zeroize instance to scalar and point byte representations
- SecretPoint and SecretScalar provide methods to get secret bytes


generic-ec/src/non_zero/mod.rs
```diff
@@ -38,6 +38,43 @@ impl<E: Curve> NonZero<Point<E>> {
         // zero value is not accessible by anyone
         CtOption::new(Self::new_unchecked(point), is_non_zero)
     }
+
+    /// Convert this value into a `NonZero<SecretPoint<E>>`. You should do this at the end
+    /// of computations that produce a secret, like a key exchange
+    pub fn into_secret(self) -> NonZero<SecretPoint<E>> {
```
survived, line 45: You should probably add `#[inline(always)]` here as well. If a separate function is called, a copy of secret point will be left on stack, we will not be able to clear it up

generic-ec/src/secret_scalar/mod.rs
```diff
@@ -61,6 +63,20 @@ impl<E: Curve> SecretScalar<E> {
         Self::random(&mut rng)
     }
 
+    /// Encodes scalar as bytes in big-endian order
+    #[cfg(feature = "alloc")]
+    pub fn to_be_bytes(&self) -> alloc::boxed::Box<zeroize::Zeroizing<EncodedScalar<E>>> {
+        let bytes = zeroize::Zeroizing::new(self.as_ref().to_be_bytes());
+        alloc::boxed::Box::new(bytes)
```
survived: here, the value of `bytes` is semantically moved into `Box`, but a value is left on stack. It'd be better to clean up the bytes on stack afterwards. Something like:

```rust
let mut bytes = self.as_ref().to_be_bytes();
let boxed_bytes = alloc::boxed::Box::new(zeroize::Zeroizing::new(bytes));
zeroize::Zeroize::zeroize(bytes.as_mut());
boxed_bytes
```

generic-ec/src/secret_scalar/mod.rs
```diff
@@ -61,6 +63,20 @@ impl<E: Curve> SecretScalar<E> {
         Self::random(&mut rng)
     }
 
+    /// Encodes scalar as bytes in big-endian order
+    #[cfg(feature = "alloc")]
+    pub fn to_be_bytes(&self) -> alloc::boxed::Box<zeroize::Zeroizing<EncodedScalar<E>>> {
```
survived: Up until now, we've been providing the same API whether `alloc` feature is used or not. This method can be provided to non-alloc targets by having

```rust
#[cfg(feature = "alloc")]
pub struct EncodedSecretScalar<E: Curve>(alloc::boxed::Box<zeroize::Zeroizing<EncodedScalar<E>>>);
#[cfg(not(feature = "alloc"))]
pub struct EncodedSecretScalar<E: Curve>(zeroize::Zeroizing<EncodedScalar<E>>);
```

similar to how secret scalar/point are defined

generic-ec/src/non_zero/mod.rs
survived, line 45: can you also add a similar method to `NonZero<Scalar<E>>` and `Scalar<E>` ?

generic-ec/src/secret_scalar/mod.rs
maurges: I specifically took care that the value left on stack is `Zeroizing` so that it will get cleared after the function quits. It may be still that the intermediary value is left in the memory though, which maybe has less chance to happen with your approach

generic-ec/src/secret_scalar/mod.rs
maurges: But then the api is not the same, you can't reuse the code between alloc and non-alloc. I could come up with a wrapper struct that exposes the same AsRef whether the underlying bytes are boxed or not. Or do you have a better idea?

generic-ec/src/non_zero/mod.rs
maurges: Right, because this one I mindlessly copied from SecretScalar and the other one I wrote myself. I'll add the inline to all occurings of this method

generic-ec/src/secret_scalar/mod.rs
survived: Wdym? If `SecretScalar::to_be_bytes()` returns `EncodedSecretScalar` regardless of `alloc` feature, we will have exactly the same API

survived: as long as `EncodedSecretScalar` exposes the same methods

survived: I don't see any better solution than a wrapper struct

generic-ec/src/secret_scalar/mod.rs
survived: In your code, `Zeroizing` does not clear up the value left on stack, it will only clear up the value on the heap

survived: `Zeroizing` clears up data on drop, there's no drop happening in this function

generic-ec/src/secret_scalar/mod.rs
maurges: ..I don't know how I read your message yesterday, but yes, what you wrote there is also what came to my mind first

generic-ec/src/non_zero/mod.rs
maurges: `NonZero<Scalar<E>>` already has it. Added to scalar

survived: > Added to scalar

You didn't push it though

maurges: Github doesn't show this chunk as outdated, but the inline is there

generic-ec/src/encoded.rs
```diff
@@ -160,3 +169,161 @@ impl<E: Curve> Zeroize for EncodedScalar<E> {
...
+        core::mem::swap(&mut **value_mut, x);
+        x.zeroize();
+        value_on_heap
+    }
+
+    pub(super) fn inner<T>(x: Secret<T>) -> T
+    where
+        T: zeroize::Zeroize + Clone,
+    {
+        (**x).clone()
```
survived: The function signature suggest that it takes ownership of `x` and retrieves the wrapped value, where as in reality it just clones the wrapped value.

* I think it should accept `x: &Secret<T>`. It's not only clearer, but also more efficient, as we don't need to increment/decrement `Arc`'s reference counter
* Another name could be a better fit. E.g. `clone_inner`?

generic-ec/src/encoded.rs
```diff
@@ -160,3 +169,161 @@ impl<E: Curve> Zeroize for EncodedScalar<E> {
...
     }
 }
+
+// Internals for both SecretEncodedPoint and SecretEncodingScalar depending on
+// presence of alloc
+#[cfg(feature = "alloc")]
+mod imp {
+    use alloc::sync::Arc;
+
+    pub(super) type Secret<T> = Arc<zeroize::Zeroizing<T>>;
```
survived: That's a cool abstraction! I wonder if it could be even reused for `SecretScalar` and `SecretPoint`, as they also both store `Arc<zeroize::Zeroizing<T>>` inside? Prob this module would need to get a few more methods (e.g. `fn as_ref(x: &Secret<T>) -> &T`)

generic-ec/src/encoded.rs
```diff
@@ -160,3 +169,161 @@ impl<E: Curve> Zeroize for EncodedScalar<E> {
...
+    }
+
+    /// Obtain the reference to the encoded bytes. This bypasses all the secrecy
+    /// guarantees, so be careful when handling them
+    pub fn as_bytes(&self) -> &[u8] {
+        self.0.as_bytes()
+    }
+
+    /// Convert the representation into not-secret, dropping all guarantees
+    pub fn into_not_secret(self) -> EncodedPoint<E> {
```
survived: I like the naming. Probably, `SecretScalar::as_ref()` should be renamed into `as_not_secret()` to be clearer. 

That would break the API, but we could deprecate `as_ref` and provide a new `as_not_secret` as the suggested replacement.

generic-ec/src/encoded.rs
```diff
@@ -160,3 +169,161 @@ impl<E: Curve> Zeroize for EncodedScalar<E> {
...
+    fn deref(&self) -> &[u8] {
+        self.as_bytes()
+    }
+}
+
+impl<E: Curve> AsRef<[u8]> for EncodedSecretPoint<E> {
+    fn as_ref(&self) -> &[u8] {
+        self.as_bytes()
+    }
+}
```
survived: Should we remove `Deref` and `AsRef<[u8]>` to force to use `as_bytes()`? The latter has a clear documentation that states:

```rust
    /// Obtain the reference to the encoded bytes. This bypasses all the secrecy
    /// guarantees, so be careful when handling them
```

survived: Especially `Deref`, cuz compiler does dereferencing automatically

generic-ec/src/encoded.rs
```diff
@@ -160,3 +169,161 @@ impl<E: Curve> Zeroize for EncodedScalar<E> {
...
+impl<E: Curve> EncodedSecretPoint<E> {
+    /// Wrap a non-secret representation
+    #[inline(always)]
+    pub fn new(mut point: EncodedPoint<E>) -> Self {
+        Self(imp::new(&mut point))
+    }
+
+    /// Obtain the reference to the encoded bytes. This bypasses all the secrecy
+    /// guarantees, so be careful when handling them
+    pub fn as_bytes(&self) -> &[u8] {
```
survived: in the spirit of `into_not_secret` method, I'm inclined to propose to rename this one to `as_not_secret_bytes`, though that becomes quite long (but maybe it's justified?)

generic-ec/src/encoded.rs
```diff
@@ -160,3 +169,161 @@ impl<E: Curve> Zeroize for EncodedScalar<E> {
...
+    fn deref(&self) -> &[u8] {
+        self.as_bytes()
+    }
+}
+
+impl<E: Curve> AsRef<[u8]> for EncodedSecretScalar<E> {
+    fn as_ref(&self) -> &[u8] {
+        self.as_bytes()
+    }
+}
```
survived: Same suggestions as for `Deref`/`AsRef<[u8]>` for `EncodedSecretPoint`

generic-ec/src/encoded.rs
survived: Ah I just realized that we can't do that, since we rely a lot on that `&SecretScalar` can be passed to a function that accepts `x: impl AsRef<Scalar<E>>`...

generic-ec/src/encoded.rs
maurges: When I was implementing this, I was thinking about the non-Arc case: you take the secret value, take it out of the Zeroizing, and obtain a non-secret. But then I realized that you can't into_inner the Zeroizing struct, and can only clone from within, but I was commited to the function idea. Now I need to reflect on what I want from it.

generic-ec/src/encoded.rs
maurges: I can add the method with this name at least

generic-ec/src/encoded.rs
maurges: Maybe slight shortening to `as_nonsecret_bytes`? I was initially debating the method name between `into_not_secret` and `into_nonsecret`

generic-ec/src/encoded.rs
maurges: We can, you give a good reasoning. This whole PR was conceived for the ecies crate, and it only requires as_bytes, everything else I added for completion with non-secret

generic-ec/src/encoded.rs
maurges: It seems that it can, I will refactor

generic-ec/src/encoded.rs
maurges: Though its implementation will be as stupid as for the encoded one, with clone, because as I see I can't convert out of Zeroizing.

Unless we want to use unsafe transmute, which will work because Zeroizing is guaranteed to have the same representation

generic-ec/src/secret.rs
```diff
@@ -0,0 +1,55 @@
+//! Internals for both Secret structs, like `SecretScalar` and `SecretEncodedPoint`. They have a different representation based on the presence of `alloc`: either a reference-counted nonmoving one, or a moving value
```
survived: (there are more than 2 secret structs)

```suggestion
//! Internals for Secret structs, like `SecretScalar` and `SecretEncodedPoint`. They have a different representation based on the presence of `alloc`: either a reference-counted nonmoving one, or a moving value
```

generic-ec/src/encoded.rs
survived: > Though its implementation will be as stupid as for the encoded one, with clone, because as I see I can't convert out of Zeroizing.

Why? You can reference inner data (via Deref on Zeroizing), you don't need to clone it

generic-ec/src/secret.rs
```diff
@@ -0,0 +1,55 @@
...
+        T: zeroize::Zeroize + Default + Clone,
+    {
+        let mut value_on_heap = Arc::<zeroize::Zeroizing<T>>::default();
+        let value_mut = Arc::make_mut(&mut value_on_heap);
+        core::mem::swap(&mut **value_mut, x);
+        x.zeroize();
+        value_on_heap
+    }
+
+    pub fn inner<T>(x: Secret<T>) -> T
```
survived: `clone_inner` feels like better name for it

generic-ec/src/encoded.rs
survived: `as_nonsecret_bytes` sounds good to me!

generic-ec/src/encoded.rs
survived: I think this method would be a great addition. It would replicate existing `AsRef` implementation. We might consider in the future removing `AsRef` to force users using explicit method with "dangerous" name.

generic-ec/src/secret.rs
maurges: I think I'll just remove this, and put `inner_ref().clone()` where necessary (which turned out nowhere)

generic-ec/src/encoded.rs
maurges: Added `as_nonsecret` which returns a reference instead, and used that in AsRef

generic-ec/src/secret.rs
survived: seems to be not working!

maurges: Yeah, why the hell does `cargo check` not check this place, but compiling for wasm does?

maurges: Wait, cargo check in CI does the check, but when I run the exact same command locally, I get no errors. What the shit

maurges: And it's also flaky...........

survived (dismissed):
Looks great! Do you want to cut a release with those changes? Then we need a version bump + changelog update

survived (approved):

PR #94, opened 2026-07-06
Fix cargo-rdme CI

maurges:
Pin the crate to the older version which worked fine and did not have this problem. Also bump the cargo-install package as it had a deprecation warning


survived (approved):
🙏

